Issue #39, opened 2022-05-04
Make possible to use without alloc/heap


leighmcculloch:
Similar to #38 there's value in us providing use of this lib in situations where `alloc` isn't available or the heap isn't ideal to use. 

This isn't urgent to do now, but something that would be nice to do.

leighmcculloch:
Two challenges with this:

- Recursive types like `ClaimPredicate` currently use the `Box` type. We might be able to use the `heapless` pooled `Box` type instead. Or maybe we drop any type containing recursion, or replace the use of the type with some other type like `ClaimPredicateNoRecursion`. Or maybe we generate unique types up to some depth where the final depth has nothing.

- Variable length arrays are `Vec` types under the hood. We could allocate fixed length arrays at the max length of the variable length type and use them instead. That's pretty simple, and `heapless` has a type we can use for that too. There is one additional challenge that our XDR contains unbounded variable length arrays and these realistically can't be allocated at u32::MAX. We'd need to provide some max capacity and hope for the best. Alternatively we just drop any type with a variable length field.

leighmcculloch:
As a follow on from https://github.com/stellar/rs-stellar-xdr/pull/44, this issue is being repurposed to change the `alloc` feature so it is possible to use the default global allocator, or provide your own allocator.

PR #44, opened 2022-05-13
Add support for no-alloc using static lifetime ref Box and Vec types

leighmcculloch:
### What

Add some super limited support for type construction in no-alloc envs using static lifetime ref `Box` and `Vec` types, used whenever the `alloc` feature is disabled. The `alloc` feature is dependent on the `alloc` crate's `Box` and `Vec`, and is enabled whenever `std` is enabled.

This results in three tiers of capabilities:
- `std`, type construction using the default allocator for recursive types (that use `Box`) and `Vec`s, and encoding/decoding.
- `alloc`, type construction using the default allocator for recursive types (that use `Box`) and `Vec`s.
- Otherwise, type construction using static lifetime refs for recursive types and `Vec`s.

### Why

This crate has good support for `no_std` environments, but not for environments that require no allocator. That's because the XDR types for the Stellar protocol are recursive, and so we can't easily drop-in stack based Vec types like heapless. There are some uses of this lib in no alloc environments that involve using the types for structured programming without us actually needing encoding/decoding. This PR makes that possible without changing the interface of passing an allocator around. Ideally in #39 this is revisited and an allocator is passed down instead.

### Known limitations

These changes are temporary, and will be replaced when #39 is addressed.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,7 @@ jobs:
     needs: [fmt, build, test]
     runs-on: ubuntu-latest
     steps:
-    - if: failure()
+    - if: contains(needs.*.result, 'failure')
       run: exit 1
 
   fmt:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,7 +24,6 @@ jobs:
     - uses: actions/checkout@v3
     - run: rustup update stable && rustup default stable
     - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --check "{}"
-    - run: exit 1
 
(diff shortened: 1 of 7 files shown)


.github/workflows/rust.yml
```diff
@@ -10,6 +10,14 @@ env:
...
 jobs:
 
+  complete:
+    if: always()
+    needs: [fmt, build, test]
+    runs-on: ubuntu-latest
+    steps:
+    - if: contains(needs.*.result, 'failure')
+      run: exit 1
+
```
leighmcculloch, line 20: Updates the builds for https://github.com/stellar/terraform/pull/804.

PR #70, opened 2022-06-03
Makefile fixes

graydon:
Couple fixes to the makefiles. I'm assuming you're running docker on windows or macos -- somewhere in a VM -- where it runs as a local user. In linux it runs as root and files it writes are not user-modifiable. So I added a chown.

The other change catches I think a typo during the migration to curr and next -- weren't deleting the .x files which meant they didn't get re-downloaded and regenerated on clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,6 @@ src/curr.rs: $(XDR_FILES_LOCAL_CURR)
 		gem specific_install https://github.com/stellar/xdrgen.git -b master && \
 		xdrgen --language rust --namespace curr --output src/ $^ \
 		'
-	sudo chown $$USER $@
 	rustfmt $@
 
 src/next.rs: $(XDR_FILES_LOCAL_NEXT)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -57,15 +57,17 @@ src/curr.rs: $(XDR_FILES_LOCAL_CURR)
 		gem specific_install https://github.com/stellar/xdrgen.git -b master && \
 		xdrgen --language rust --namespace curr --output src/ $^ \
 		'
-	rustfmt src/curr.rs
+	sudo chown $$USER $@
```
leighmcculloch, line 60: I want to fix this issue by changing the docker command, so that we aren't running sudo on the host. Let me circle back on that.

leighmcculloch (commented):
Looks good, but I want to fix the sudo issue another way.

Makefile
graydon: It appears to be phenomenally complex to make this work. https://jtreminio.com/blog/running-docker-containers-as-current-host-user/

I'm a bit surprised, but not really surprised.

leighmcculloch: This should do it. I tested it on Ubuntu. Let me know if it doesn't fix it for you. https://github.com/stellar/rs-stellar-xdr/pull/71

PR #73, opened 2022-06-14
CAP-52 XDR

jonjove:
### What

Add the XDR for CAP-52, specifically `InvokeHostFunctionOp`.

### Why

Required to support CAP-52 in stellar/rs-stellar-contract-env and stellar/stellar-core.

### Known limitations

N/A

### DIFF
--- a/src/next.rs
+++ b/src/next.rs
@@ -8926,7 +8926,8 @@ impl WriteXdr for LedgerFootprint {
 //        CLAWBACK_CLAIMABLE_BALANCE = 20,
 //        SET_TRUST_LINE_FLAGS = 21,
 //        LIQUIDITY_POOL_DEPOSIT = 22,
-//        LIQUIDITY_POOL_WITHDRAW = 23
+//        LIQUIDITY_POOL_WITHDRAW = 23,
+//        INVOKE_HOST_FUNCTION = 24
 //    };
 //
 // enum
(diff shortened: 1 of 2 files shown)


jonjove:
@leighmcculloch it took me a while to get all the PRs open, but it corresponds to changes in https://github.com/stellar/stellar-core/pull/3450 and https://github.com/stellar/rs-stellar-contract-env/pull/112. I don't know how we want to sequence these merges.

leighmcculloch:
@jonjove I think if you want to be strict about it, you should break up stellar/stellar-core#3450 to be two PRs, one that modifies the XDR first, and then one that makes any changes that are dependent on the `stellar-env-host` crate. Then the process would be:
- Modify XDR.
- Update rs-stellar-xdr.
- Update rs-stellar-env-host.
- Update stellar-core.

We may as well have the XDR in its own repo. cc @MonsieurNicolas 

@graydon How did you do this recently?

leighmcculloch:
Until we figure this out I think it's fine if you merge this just before merging the stellar-core PR, assuming the .x files are identical.

leighmcculloch (approved):

graydon (changes requested):
I left a comment in the accompanying core XDR change that I think the return from the function should have an SCVal and so this would need to be regenerated to match if so. Otherwise ok.

jonjove:
@graydon I left a long [comment](https://github.com/stellar/stellar-core/pull/3450#discussion_r898041956) explaining why I don't think we should have the `SCVal`.

PR #76, opened 2022-06-16
Add contract spec XDR

leighmcculloch:
### What

Add contract spec XDR.

### Why

We need a serialized form of a contract interface.

Close https://github.com/stellar/rs-stellar-contract-sdk/issues/115
Related to https://github.com/stellar/rs-stellar-contract-sdk/pull/140
Related to stellar/stellar-contract-cli#17

### Known limitations

[TODO or N/A]


leighmcculloch:
I'm merging the spec xdr because there are enough components using it in the sdk + cli that it is functional, however it is still very much in the early design, and it will likely change to better support and align with https://github.com/stellar/rs-stellar-contract-env/issues/139.

PR #85, opened 2022-06-22
Expand `ScStatus` type, include various error codes

jayz22:
### What

### Why

This is needed for host to support fallible version of contract calling -- returning a substantial error code instead of trapping, which allows the calling code to continue running.

### Known limitations

[TODO or N/A]


xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    BITSET_TOO_MANY_BITS = 10,
+    STATUS_UNKNOWN = 11
+};
+
+enum SCHostObjErrorCode
+{
+    HOST_OBJECT_UNKNOWN_ERROR = 0,
+    UNKNOWN_HOST_OBJECT_REFERENCE = 1,
+    UNEXPECTED_HOST_OBJECT_TYPE = 2,
+    OBJECT_HANDLE_EXCEEDS_U32_MAX = 3,
```
leighmcculloch, line 105: How can this error occur given that object handles are bound to u32 max?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    ACCESSING_HOST_OBJECT_OUT_OF_BOUND = 4,
+    VEC_INDEX_OUT_OF_BOUND = 5,
+    VEC_INDEX_OVERFLOW = 6,
```
leighmcculloch, line 108: What's the significance of an index overflow vs out of bound error for a vec?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    STATUS_UNKNOWN = 11
+};
+
+enum SCHostObjErrorCode
+{
+    HOST_OBJECT_UNKNOWN_ERROR = 0,
+    UNKNOWN_HOST_OBJECT_REFERENCE = 1,
+    UNEXPECTED_HOST_OBJECT_TYPE = 2,
+    OBJECT_HANDLE_EXCEEDS_U32_MAX = 3,
+    ACCESSING_HOST_OBJECT_OUT_OF_BOUND = 4,
```
leighmcculloch, line 106: What does this error indicate?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    VEC_INDEX_OUT_OF_BOUND = 5,
+    VEC_INDEX_OVERFLOW = 6,
+    VEC_VALUE_NOT_EXIST = 7,
```
leighmcculloch, line 109: When does a vec value not exist?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    INVALID_CONTRACT_HASH = 8
```
leighmcculloch, line 110: Given that contract hashes are bytes and other than length probably have no validation, when are they invalid?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+};
+
+enum SCHostFnErrorCode
+{
+    HOST_FN_UNKNOWN_ERROR = 0,
+    UNEXPECTED_HOST_FUNCTION_ACTION = 1,
+    UNEXPECTED_ARGS = 2,
+    INVALID_ARGS = 3,
+    WRONG_INPUT_ARG_TYPE = 4
```
leighmcculloch, line 119: What's the distinction between these arg related errors? Is `UNEXPECTED_` indicative of too many arguments provided? Which error occurs for too few arguments provided? When are args invalid vs the wrong type?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+};
+
+enum SCHostStorageErrorCode
+{
+    HOST_STORAGE_UNKNOWN_ERROR = 0,
+    EXPECT_CONTRACT_DATA = 1,
+    READWRITE_ACCESS_TO_READONLY_ENTRY = 2,
+    ACCESS_TO_UNKNOWN_ENTRY = 3,
+    MISSING_KEY_IN_GET = 4,
+    GET_ON_DELETED_KEY = 5
```
leighmcculloch, line 129: What's the significance of a get on deleted key vs access to unknown entry?

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    WASMI_TRAP_INVALID_CONVERSION_TO_INT = 14,
+    WASMI_TRAP_STACK_OVERFLOW = 15,
+    WASMI_TRAP_UNEXPECTED_SIGNATURE = 16
+};
+
+enum SCUnknownErrorCode
+{
+    UNKNOWN_GENERAL_ERROR = 0,
+    UNKNOWN_XDR_ERROR = 1,
+    UNKNOWN_PARITY_WASMI_ELEMENTS_ERROR = 2
```
leighmcculloch, line 162: This will produce better Rust code as the prefix will be omitted, it also helps to namespace the values.
```suggestion
    UNKNOWN_ERROR_GENERAL = 0,
    UNKNOWN_ERROR_XDR = 1,
    UNKNOWN_ERROR_PARITY_WASMI_ELEMENTS = 2
```

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+
+enum SCHostContextErrorCode
+{
+    HOST_CONTEXT_UNKNOWN_ERROR = 0,
+    NO_CONTRACT_RUNNING = 1
+};
+
+enum SCVmErrorCode {
+    WASMI_UNKNOWN_ERROR = 0,
+    WASMI_VALIDATION_ERROR = 1,
```
leighmcculloch, line 140: This will produce better Rust code as the prefix will be omitted, it also helps to namespace the values.
```suggestion
    WASMI_ERROR_UNKNOWN = 0,
    WASMI_ERROR_VALIDATION = 1,
```

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    SST_HOST_FUNCTION_ERROR = 4,
+    SST_HOST_STORAGE_ERROR = 5,
+    SST_HOST_CONTEXT_ERROR = 6,
+    SST_VM_ERROR = 7
     // TODO: add more
 };
 
+enum SCHostValErrorCode
+{
+    HOST_VALUE_UNKNOWN_ERROR = 0,
```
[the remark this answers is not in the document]
jayz22: Updated.

xdr/next/Stellar-contract.x
jayz22: Updated.

xdr/next/Stellar-contract.x
jayz22: Updated.

xdr/next/Stellar-contract.x
jayz22: Internally when a new host object is added, it gets pushed into `Vec<HostObj>` stored in the `host`. The handle is the length of the Vec, which may exceed u32 max. This error check ensures it doesn't, then we cast the handle to `u32`. See [code ref](https://github.com/stellar/rs-stellar-contract-env/blob/main/stellar-contract-env-host/src/host.rs#L443-L445).

jayz22: Updated it to `HOST_OBJECT_OBJECT_COUNT_EXCEEDS_U32_MAX`.

xdr/next/Stellar-contract.x
jayz22: Removed. It was outdated.

xdr/next/Stellar-contract.x
jayz22: You are right. Changed to `HOST_OBJECT_CONTRACT_HASH_WRONG_LENGTH`.

xdr/next/Stellar-contract.x
jayz22: Changed it to `HOST_OBJECT_OBJECT_NOT_EXIST`. This happens when the index provided exceeds the length of the hostObj vec, thus trying to access object that does not exist.

xdr/next/Stellar-contract.x
jayz22: Changed to `HOST_OBJECT_VEC_ELEMENT_NOT_EXIST`.

xdr/next/Stellar-contract.x
jayz22: I changed them to
```
HOST_FN_INPUT_ARGS_WRONG_LENGTH
HOST_FN_INPUT_ARGS_WRONG_TYPE
HOST_FN_INPUT_ARGS_INVALID
```
The `HOST_FN_INPUT_ARGS_INVALID` means the args fails some validation check. E.g. in `vec_slice(v: obj, i:u32, l:32)`, `i + j`  exceeds `u32::max`.

jayz22:
Thanks @leighmcculloch for reviewing. As I work through your suggestions, I realize many of these error code names were direct literal translation of the error messages in the code, which is not the most intuitive when taken out of context. So it's definitely helpful!
In terms of merging logistics, do I need to create a separate PR on stellar-core with the exact same XDR changes and get that merged first, before merging this?

xdr/next/Stellar-contract.x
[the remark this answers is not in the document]
jayz22: > xdrgen that generates the Rust code for rs-stellar-xdr will omit prefixes that are shared 
Thanks, good to know

leighmcculloch:
>In terms of merging logistics, do I need to create a separate PR on stellar-core with the exact same XDR changes and get that merged first, before merging this?

Merge it here first, just try and merge it and the stellar-core change same day if you can.

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,114 @@ enum SCStatic
...
+    WASMI_ERROR_VALUE = 7,
+    WASMI_TRAP_UNREACHABLE = 8,
+    WASMI_TRAP_MEMORY_ACCESS_OUT_OF_BOUNDS = 9,
+    WASMI_TRAP_TABLE_ACCESS_OUT_OF_BOUNDS = 10,
+    WASMI_TRAP_ELEM_UNINITIALIZED = 11,
+    WASMI_TRAP_DIVISION_BY_ZERO = 12,
+    WASMI_TRAP_INTEGER_OVERFLOW = 13,
+    WASMI_TRAP_INVALID_CONVERSION_TO_INT = 14,
+    WASMI_TRAP_STACK_OVERFLOW = 15,
+    WASMI_TRAP_UNEXPECTED_SIGNATURE = 16
```
leighmcculloch, line 154: Could we normalize the prefix on these? Because some are WASMI_ERROR_ and others are WASMI_TRAP_ in generated code we end up with lots of ERROR_ and TRAP_.

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,114 @@ enum SCStatic
...
+};
+
+enum SCUnknownErrorCode
+{
+    UNKNOWN_ERROR_GENERAL = 0,
+    UNKNOWN_ERROR_XDR = 1,
+    UNKNOWN_ERROR_PARITY_WASMI_ELEMENTS = 2
```
leighmcculloch, line 161: It's interesting that we have PARITY_WASMI and WASMI in general in the protocol. Aren't these terms implementation details? cc @graydon if you have a different opinion?

xdr/next/Stellar-contract.x
leighmcculloch: Hmm. Does this mean that behavior changes after the first time a LE has existed, but not after any subsequent time it exists? What's the use case of distinguishing between these situations?

xdr/next/Stellar-contract.x
leighmcculloch: What does it mean that a vec element doesn't exist though? Is that the same thing as index out of bounds?

xdr/next/Stellar-contract.x
leighmcculloch: >This happens when the index provided exceeds the length of the hostObj vec

Is this the same thing as index out of bounds on a vec?

leighmcculloch: Oh, I think I get it. This is when the host object handle doesn't reference an object.

leighmcculloch:
Let's just start merging these xdr changes rather than holding them and trying to coordinate them with stellar-core. I think it's going to be easier to merge early here and maintain some discrepancy than it will be to always be trying to align the merges.

xdr/next/Stellar-contract.x
jayz22: These currently happens in `vec_front`, `vec_back` and `vec_pop`, when there is no element in the vec. Should I rename it to `HOST_OBJECT_VEC_IS_EMPTY` instead?

xdr/next/Stellar-contract.x
jayz22: The reason I maintain these prefix is because this is how it's structured in WASMI, where there the `WASMI::Error::*` contains the `WASMI_ERROR_*` matching types, plus a `WASMI::Error::Trap::*` with sub-types for each trap. I'll rename the `WASMI_TRAP_` to `WASMI_ERROR_TRAP_` so that the `Error` prefix will not show up in the generated type.

xdr/next/Stellar-contract.x
jayz22: Hmm, I have no preference on this. `HostError::ParityWasmiElements` current existed in the code. I'm fine with removing it for now if that is implementation detail, or keep it for more clarify.

xdr/next/Stellar-contract.x
```diff
@@ -71,16 +71,115 @@ enum SCStatic
...
+    HOST_VALUE_UNKNOWN_ERROR = 0,
+    RESERVED_TAG_VALUE = 1,
+    UNEXPECTED_VAL_TYPE = 2,
+    U63_OUT_OF_RANGE = 3,
+    U32_OUT_OF_RANGE = 4,
+    STATIC_UNKNOWN = 5,
+    MISSING_OBJECT = 6,
+    SYMBOL_TOO_LONG = 7,
+    SYMBOL_BAD_CHAR = 8,
+    SYMBOL_CONTAINS_NON_UTF8 = 9,
```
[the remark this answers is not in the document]
jayz22: We can still have "bad char" and "non-utf8" errors if we store in i64 representation?

jayz22:
> Let's just start merging these xdr changes rather than holding them and trying to coordinate them with stellar-core. 

Sounds good, I've marked it ready for review.

xdr/next/Stellar-contract.x
leighmcculloch: I would expect the error to be the same that the index is out of bound. Distinguishing between these cases feels like an implementation details is leaking into the protocol. If we need a unique error for some reason then "vec is empty" is clearer.

xdr/next/Stellar-contract.x
leighmcculloch: Interesting. If trap errors are a type of error, maybe we should treat them the same in the protocol and make the type of trap a subfield? I don't feel strongly but just pointing out we could do that.

xdr/next/Stellar-contract.x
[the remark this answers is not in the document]
leighmcculloch: Looking at https://github.com/stellar/rs-stellar-contract-env/blob/a2e95d91d81d6cf79250e65f906cb4faa715ecca/stellar-contract-env-common/src/symbol.rs#L213 I don't see a case where the i64 version can write when decoding.
What am I missing?

xdr/next/Stellar-contract.x
jayz22: Ok, I've removed the `HOST_OBJECT_VEC_IS_EMPTY` case. Will just treat it same as "out of bound".

xdr/next/Stellar-contract.x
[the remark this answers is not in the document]
jayz22: Ah I see, you are right about the i64 case. 
But I think there is still good use cases for converting from a string? For example encoding some string as symbol to be passed around (between host and contract or between contracts). 
I don't have a perfect example right now, but one thing I'm also looking at is the logging part of things where a string need to be converted to a symbol and logged.

xdr/next/Stellar-contract.x
jayz22: > It's interesting that we have PARITY_WASMI and WASMI in general in the protocol. 

We do use PARITY_WASMI module separately from WASMI, on things like deserializing a wasm binary code (@graydon can advise more on the exact usages). They potentially emit error which we then wrap into `HostError`, which I then assign to an error status code. I could relocate the PARITY_WASMI error to under WASMI's category , but I don't feel that is accurate.

xdr/next/Stellar-contract.x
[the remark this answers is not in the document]
graydon: I wanted to not burden all the non-smart-contract-guest-language client SDKs (C#, JS, etc.) with having to learn to read and write the bit-packing encoding of symbol -- I wanted to keep it private between host and guest languages.

xdr/next/Stellar-contract.x
graydon: I would prefer not to put PARITY_WASMI nor WASMI in the protocol -- there's still a chance going forward we may change VMs, and the trap types are fairly likely to be similar if not identical between VMs.

xdr/next/Stellar-contract.x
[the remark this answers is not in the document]
leighmcculloch: Ah, makes sense. A few extra errors are a small price to pay for this convenience. Thanks for pointing that out. 👍🏻

graydon (approved):

Issue #80, opened 2022-06-17
Add a step to CI that rebuilds the generated code from the .x files in the repo


leighmcculloch:
Add a step to CI that rebuilds the generated code from the .x files in the repo and verifies that they're up-to-date.

PR #87, opened 2022-06-23
Add XDR types bigint, hash, publickey

jayz22:
### What

[TODO: Short statement about what is changing.]

### Why

CAP-51 host functions.

### Known limitations

[TODO or N/A]

PR #91, opened 2022-06-24
Check generated code matches .x files in CI

leighmcculloch:
### What

Check generated code matches .x files in CI.

### Why

It's easy to miss generating the latest code.

Close https://github.com/stellar/rs-stellar-xdr/issues/80

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,6 +18,14 @@ jobs:
     - if: contains(needs.*.result, 'failure')
       run: exit 1
 
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


leighmcculloch:
@jayz22 #87 was merged but turned out the code wasn't quite consistent. @jonjove had experienced the same thing recently. There hasn't actually been a check that code is fully up-to-date, and so it is super easy for that to happen, so this PR tries to fix that.

jayz22:
> @jayz22 #87 was merged but turned out the code wasn't quite consistent. @jonjove had experienced the same thing recently. There hasn't actually been a check that code is fully up-to-date, and so it is super easy for that to happen, so this PR tries to fix that.

Ah I see, thanks for the check.

PR #99, opened 2022-07-13
Serde support

sisuresh:
Ran xdrgen, which now includes a serde feature.


leighmcculloch:
The generate error is because `curr.rs` also needs regenerating. Running `make clean build test` should do the job.

PR #102, opened 2022-07-14
Make serde dependency optional

leighmcculloch:
### What
Make `serde` dependency optional.

### Why
The `serde` dependency shouldn't be included in the dependency tree unless the `serde` feature is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,18 +28,18 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.138"
+version = "1.0.139"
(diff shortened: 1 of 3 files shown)


leighmcculloch:
This change includes a TODO to remove the `--optional-deps` usage once weak dependencies are recognized by cargo-hack https://github.com/taiki-e/cargo-hack/issues/147 and once the bug with first-class features that have the same name as an optional dependency not being seen as a feature to test is addressed https://github.com/taiki-e/cargo-hack/issues/153.

PR #112, opened 2022-07-19
Add Validate impl for ScMap and Map ScVal variant that can be invalid

### LINKED ISSUE #113: Test Fwk: Add ScVal validation trait and fns (ScMap)
Follow up to https://github.com/stellar/rs-stellar-xdr/issues/106

leighmcculloch:
### What

Add `Validate` trait and impl for `ScMap` and `ScVal::Map` variant that can be invalid regardless of any other context.

### Why

Facilitate users of the XDR validating that the ScVal types are valid. There are likely to be multiple places that we'll need to validate these XDR values. For map types this is especially easy to mess up, and will be one thing in the toolbox that we use to prevent footguns so that users do not see signature failures for transactions that they submit a misordered map for that they signed for.

Follow up to https://github.com/stellar/rs-stellar-xdr/pull/111
Close #113

### Known limitations

[TODO or N/A]


graydon:
@leighmcculloch I've pushed a little more here to cover the "construct a sorted map" scenario and tidied the comment. Hope it covers what you wanted! Let me know.

leighmcculloch (commented):
✅ Thanks @graydon !

src/scmap.rs
```diff
@@ -0,0 +1,68 @@
...
+#![allow(clippy::missing_errors_doc)]
+
+use crate::{ScMap, ScMapEntry, ScVal, Validate};
+extern crate alloc;
+use alloc::vec::Vec;
+
+impl ScMap {
+    pub fn sorted_from_entries<I: Iterator<Item = ScMapEntry>>(entries: I) -> Result<ScMap, ()> {
+        let mut v: Vec<ScMapEntry> = entries.collect();
+        v.sort_by(|a, b| a.key.cmp(&b.key));
```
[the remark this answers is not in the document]
leighmcculloch:
```suggestion
        // TODO: Add tests that prove order consistency of ScVal with RawVal. https://github.com/stellar/rs-stellar-xdr/issues/117
        v.sort_by(|a, b| a.key.cmp(&b.key));
```

src/scmap.rs
```diff
@@ -0,0 +1,69 @@
...
+        let m = ScMap(v.try_into()?);
+        // `validate` here will further check that there are no duplicates.
+        m.validate()?;
+        Ok(m)
+    }
+
+    pub fn sorted_from_pairs<K, V, I: Iterator<Item = (K, V)>>(pairs: I) -> Result<ScMap, ()>
+    where
+        ScVal: From<K> + From<V>,
```
leighmcculloch, line 21: It might create more flexibility to make this use the Into conversion, or better yet, the TryInto conversion. Using Into instead of From will make it work for types that implement only the Into conversion. Using the Try version will make it work with types that don't offer an errorless conversion, like Vec.

graydon: Done.

src/scmap.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn scmap_from_pairs() -> Result<(), ()> {
+        let pairs: Vec<(u32, u32)> = vec![(3, 4), (5, 6), (1, 2)];
+        let scm = ScMap::sorted_from(pairs)?;
+        assert_eq!(scm.0.first().unwrap().key, 1u32.into());
+        assert_eq!(scm.0.last().unwrap().key, 5u32.into());
+        Ok(())
+    }
+
+    #[test]
+    fn scmap_from_bad_pairs() {
```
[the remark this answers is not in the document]
graydon: Done.

leighmcculloch:
Thanks @graydon. I made some more tweaks to handle keys and values that are `TryInto<ScVal>` rather than `Into<ScVal>`. I realized after we conversed on here earlier that some types don't have infallible conversions to `ScVal` and would be excluded.

src/scmap.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    pub fn sorted_from_pairs<K, V, I>(pairs: I) -> Result<ScMap, Error>
+    where
+        K: TryInto<ScVal>,
+        V: TryInto<ScVal>,
+        I: Iterator<Item = (K, V)>,
+    {
+        Self::sorted_from_entries(pairs)
+    }
```
leighmcculloch, line 32: This fn is now technically redundant because the same values can be passed directly to the entries fn. But this fn has a clearer signature that is easier to understand, so I think it is still useful so I kept it.

src/scval_validations.rs
```diff
@@ -63,39 +63,134 @@ impl Validate for ScVal {
...
-    type Error = ();
+    type Error = Error;
 
     fn validate(&self) -> Result<(), Self::Error> {
-        // TODO: Validate that the map is sorted and has no duplicates, or find
-        // a way to guarantee this to be the case.
-        todo!()
+        // Check the map is sorted by key, and there are no keys that are
+        // duplicates.
+        if self.windows(2).all(|w| w[0].key < w[1].key) {
```
jayz22, line 71: What is this `.windows` function? Can't find its definition.

leighmcculloch: The function is https://doc.rust-lang.org/std/primitive.slice.html#method.windows.

jayz22: Ah cool!

PR #132, opened 2022-08-05
Support Arbitrary

graydon:
Obviously this needs to move upstream to xdrgen (I will post a version of the minimal change needed there too) but this is a sketch of what support for `Arbitrary` looks like. Very straightforward -- the only question I think I have is whether @leighmcculloch you'd like it to be cfg'ed behind another feature or not. Argument for feature-cfg'ing is that obviously "it's more code" people might not want to compile; argument against is that "every feature makes a larger powerset".


leighmcculloch:
I commented on https://github.com/stellar/xdrgen/pull/127, let's discuss there.

graydon:
LGTM!

PR #143, opened 2022-08-24
Regenerate from stellar-xdr-next@35f1e4e

graydon:
Regenerate based on changes that landed in stellar-xdr-next and some changes that first landed here and then were ported over to stellar-xdr-next in https://github.com/stellar/stellar-xdr-next/pull/21

### DIFF
--- a/xdr/next/Stellar-contract.x
+++ b/xdr/next/Stellar-contract.x
@@ -168,17 +168,17 @@ case SST_OK:
 case SST_UNKNOWN_ERROR:
     SCUnknownErrorCode unknownCode;
 case SST_HOST_VALUE_ERROR:
-    SCHostValErrorCode errorCode;
+    SCHostValErrorCode valCode;
 case SST_HOST_OBJECT_ERROR:
-    SCHostObjErrorCode errorCode;
+    SCHostObjErrorCode objCode;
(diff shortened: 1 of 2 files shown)


xdr/next/Stellar-contract.x
```diff
@@ -168,17 +168,17 @@ case SST_OK:
...
+    SCHostFnErrorCode fnCode;
 case SST_HOST_STORAGE_ERROR:
-    SCHostStorageErrorCode errorCode;
+    SCHostStorageErrorCode storageCode;
 case SST_HOST_CONTEXT_ERROR:
-    SCHostContextErrorCode errorCode;
+    SCHostContextErrorCode contextCode;
 case SST_VM_ERROR:
-    SCVmErrorCode errorCode;
+    SCVmErrorCode vmCode;
```
jayz22, line 181: Did you manually copy these changes from `stellar-xdr-next` or were these picked up automatically?
I guess it's not clear to me is, why do we have these local files if the source of changes should come from `stellar-xdr-next`? Wouldn't it make sense for them to live in the same repo/workspace? 
When I makes changes to the `.x` files in the `stellar-xdr-next`, it would be nice for me to run the regen code to make sure no syntactical errors, which makes it feel more natural to make those changes here.

PR #188, opened 2022-11-11
Add Lib version and XDR version to library

leighmcculloch:
### What
Add XDR version to library


leighmcculloch:
Closing in favor of #189.

PR #189, opened 2022-11-12
Add Lib version and XDR version to library (submodule edition)

leighmcculloch:
### What
Add a `VERSION` const to the library that contains the version of the library, as well as the version of the XDR that the library contains.

### Why
To make it clear what XDR has been included in the library and make it easier to reconcile the use of the same XDR across multiple tools using this XDR library or other XDR libraries.


graydon (approved):
LGTM!

PR #191, opened 2022-11-21
Update to transition from BigInt to {ui}128

graydon:
Absorb changes to XDR in https://github.com/stellar/stellar-xdr/pull/57 and https://github.com/stellar/stellar-xdr/pull/58 -- removing BigInt and adding I128 and U128

### DIFF
--- a/src/scval_conversions.rs
+++ b/src/scval_conversions.rs
@@ -1,4 +1,6 @@
-use crate::{AccountId, ScMap, ScMapEntry, ScObject, ScStatic, ScStatus, ScSymbol, ScVal, ScVec};
+use crate::{
+    AccountId, Int128Parts, ScMap, ScMapEntry, ScObject, ScStatic, ScStatus, ScSymbol, ScVal, ScVec,
+};
 
 #[cfg(all(not(feature = "std"), feature = "alloc"))]
 extern crate alloc;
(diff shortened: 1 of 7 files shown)


leighmcculloch (approved):

src/scval_conversions.rs
```diff
@@ -1,15 +1,10 @@
-use crate::{
-    AccountId, ScBigInt, ScMap, ScMapEntry, ScObject, ScStatic, ScStatus, ScSymbol, ScVal, ScVec,
-};
+use crate::{AccountId, ScMap, ScMapEntry, ScObject, ScStatic, ScStatus, ScSymbol, ScVal, ScVec};
 
 #[cfg(all(not(feature = "std"), feature = "alloc"))]
 extern crate alloc;
 #[cfg(all(not(feature = "std"), feature = "alloc"))]
 use alloc::{string::String, vec, vec::Vec};
```
leighmcculloch, line 6: Can you add conversions for u128 and i128 here similar to the 64bit ones?

leighmcculloch: Why: ScVal to and from conversions are needed in the SDK for all types as there are things like message signing in tests that rely on them. Without this the types won't be usable in UDTs that require all their fields to be convertible.

leighmcculloch: See line 113+.

graydon: added

leighmcculloch: Thanks!

sisuresh: Looks like clippy has issues with the casts in the last commit.

PR #192, opened 2022-11-24
pub mod int128_helpers

graydon:
This just exposes some of the i128/u128 helper functions in the previous PR in a pub mod, so we can use them in the host.


leighmcculloch (approved):
I made the couple changes below.

PR #193, opened 2022-11-28
Update crate-git-revision to 0.0.3

leighmcculloch:
### What
Update crate-git-revision to 0.0.3.

### Why
To get the fix in crate-git-revision that stops it from invalidating builds unnecessarily.


graydon (approved):
Wonderful, thanks!

PR #209, opened 2022-12-01
Fix embedding xdr version

leighmcculloch:
### What
Change how the xdr version is embedded to store it in a file at the time of generating the code, and loading that file. Do this instead of getting the version in the build.rs.

### Why
Git and the submodules aren't part of the crate when it is packaged for crates.io, so we need to store the submodule version in a simple file before packaging occurs.


src/lib.rs
```diff
@@ -114,8 +114,8 @@ pub const VERSION: Version = Version {
     } else {
         ""
     },
-    xdr_curr: env!("XDR_CURR_VERSION"),
-    xdr_next: env!("XDR_NEXT_VERSION"),
+    xdr_curr: include_str!("../xdr/curr-version"),
+    xdr_next: include_str!("../xdr/next-version"),
```
leighmcculloch, line 118: Taking a page out of @tsachiherman's approach to implementing this in the soroban-rpc project 👏🏻

sisuresh (approved):

tsachiherman (approved):

PR #215, opened 2022-12-07
Make apt-get installs more resilient

leighmcculloch:
### What
Run apt-get update before apt-get install.

### Why
We're seeing failures which are likely to do with the apt-get repositories that are cached being out of date. It's common practice to run update before install.


paulbellamy (approved):

leighmcculloch:
Superseded by #218

PR #220, opened 2022-12-08
Add guess subcommand

leighmcculloch:
### What
Add guess subcommand that will guess what type some XDR data is.

### Why
Sometimes I don't remember what type I'm supposed to be decoding data as. This is an imperfect guesser that'll provide some suggestions.

### DIFF
--- a/src/bin/stellar-xdr/guess.rs
+++ b/src/bin/stellar-xdr/guess.rs
@@ -70,41 +70,57 @@ macro_rules! run_x {
     ($f:ident, $m:ident) => {
         fn $f(&self) -> Result<(), Error> {
             let mut f = ResetRead::new(self.file()?);
-            for v in stellar_xdr::$m::TypeVariant::VARIANTS {
+            'variants: for v in stellar_xdr::$m::TypeVariant::VARIANTS {
                 f.reset();
-                let (ok, err) = match self.input {
+                let count: usize = match self.input {
                     InputFormat::Single => stellar_xdr::$m::Type::read_xdr_to_end(v, &mut f)
-                        .map(|_| (1, 0))
-                        .unwrap_or_else(|_| (0, 1)),
+                        .ok()
+                        .map(|_| 1)
+                        .unwrap_or_default(),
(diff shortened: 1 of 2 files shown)


src/bin/stellar-xdr/guess.rs
```diff
@@ -0,0 +1,211 @@
...
+                    InputFormat::SingleBase64 => {
+                        stellar_xdr::$m::Type::read_xdr_base64_to_end(v, &mut f)
+                            .map(|_| (1, 0))
+                            .unwrap_or_else(|_| (0, 1))
+                    }
+                    InputFormat::Stream => stellar_xdr::$m::Type::read_xdr_iter(v, &mut f)
+                        .take(self.certainty)
+                        .fold((0, 0), |c, v| match v {
+                            Ok(_) => (c.0 + 1, c.1),
+                            Err(_) => (c.0, c.1 + 1),
```
leighmcculloch, line 88: These lines don't early exit on failure, and it would be better if they did. There's probably a more efficient way to do this, but I didn't want to spend too much time on it so opted for this at the moment.

src/bin/stellar-xdr/guess.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    // Input format of the XDR
+    #[arg(long, value_enum, default_value_t)]
+    input: InputFormat,
+
+    // Output format
+    #[arg(long, value_enum, default_value_t)]
+    output: OutputFormat,
+
+    /// Certainty as an arbitrary value
```
sisuresh, line 38: Is this the number of suggestions to process?

sisuresh (approved):

src/bin/stellar-xdr/guess.rs
leighmcculloch: It's the number of value reads to make on a stream, if it is a stream, to confirm they continue to decode fine before accepting that guess. I didn't write this into the comment because I wanted to keep it somewhat opaque beyond the general idea that bigger number == more confidence.

PR #221, opened 2022-12-09
Add makefile rules for regenerating from a local xdrgen

brson:
### What

When the LOCAL_XDRGEN env var is defined, `make generate` will run xdrgen from a local copy.

### Why

There isn't currently a good workflow for regenerating rs-stellar-xdr from a local copy of xdrgen.

### Known limitations

xdrgen has to be rebuilt each time it's run inside docker, so it's not fast, but good enough for limited hacking on xdrgen

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,18 @@ generate: src/curr/generated.rs xdr/curr-version src/next/generated.rs xdr/next-
 
 src/curr/generated.rs: $(sort $(wildcard xdr/curr/*.x))
 	> $@
+ifeq ($(LOCAL_XDRGEN),)
 	docker run -i --rm -v $$PWD:/wd -w /wd docker.io/library/ruby:latest /bin/bash -c '\
 		gem install specific_install -v 0.3.7 && \
 		gem specific_install https://github.com/stellar/xdrgen.git -b $(XDRGEN_VERSION) && \
 		xdrgen --language rust --namespace generated --output src/curr $^ \
 		'
+else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,23 +31,37 @@ generate: src/curr/generated.rs xdr/curr-version src/next/generated.rs xdr/next-
...
 	> $@
+ifeq ($(LOCAL_XDRGEN),)
 	docker run -i --rm -v $$PWD:/wd -w /wd docker.io/library/ruby:latest /bin/bash -c '\
 		gem install specific_install -v 0.3.7 && \
 		gem specific_install https://github.com/stellar/xdrgen.git -b $(XDRGEN_VERSION) && \
 		xdrgen --language rust --namespace generated --output src/curr $^ \
 		'
+else
+	docker run -i --rm -v $$PWD/../xdrgen:/xdrgen -v $$PWD:/wd -w /wd docker.io/library/ruby:latest /bin/bash -c '\
+		pushd /xdrgen && bundle install --deployment && rake install && popd && \
```
leighmcculloch, line 42: I pushed a commit adding `--deployment` so that the installed gems are installed inside the working directory rather than the system gem store, which will speed up the install process on repeated runs.

leighmcculloch (approved):

Makefile
leighmcculloch: This won't cause diffs in the xdrgen folder because the files are written to `/vendor/` and that directory is gitignored.

PR #243, opened 2023-04-10
Fix U/I128 and U/I256 number representation for consistent sorting

jayz22:
### What

Generate from https://github.com/stellar/stellar-xdr/pull/78 and adapted the u/i128 conversions.

### Why

### Known limitations

[TODO or N/A]


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "xdr/next"]
 	path = xdr/next
-	url = https://github.com/stellar/stellar-xdr
-	branch = next
+	url = https://github.com/jayz22/stellar-xdr
```
graydon, line 3: I've approved the associated change on stellar/stellar-xdr so you can switch this reference back

src/next/scval_conversions.rs
```diff
@@ -166,36 +168,51 @@ pub mod int128_helpers {
...
-    #[allow(clippy::inline_always, clippy::cast_possible_wrap)]
-    pub fn i128_from_u128(u: u128) -> i128 {
-        u as i128
+    #[allow(
+        clippy::inline_always,
+        clippy::cast_sign_loss,
+        clippy::cast_possible_wrap
+    )]
+    pub fn i128_from_pieces(hi: i64, lo: u64) -> i128 {
+        ((hi as u128) << 64 | u128::from(lo)) as i128
```
graydon, line 215: I don't actually know what happens when you do a direct i64-to-u128 conversion in Rust as you're doing in the first expression `(hi as u128)`. I _think_ it does a sign extending i64-to-i128, followed by a sign-change "no-op" i128-to-u128 that does not change any bits. But I'm not certain. It might also do a sign-change no-op from i64-to-u64 followed by a zero-extension u64-to-u128. Which, if that's what it does, will be wrong here. The [official semantics are here](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics) and suggest your code is doing the right thing, but I'm still not 100% certain.

My rule of thumb for making sure I don't make such errors in Rust is to do only one such action at a time: either switch signedness without changing width, or change width without changing signedness. And if you have to do both, take an extra nesting level of expressions or line of code (it all compiles away anyways) and do the two steps in two separate expressions or statements.

jayz22: Good point, I will take the extra step to make those conversions more explicit. 
(I've referred to the official doc as well, but I agree it's good to be extra-safe)

jayz22: I have changed this line, hope it's more clear now.

jayz22: (FWIW, I found the [official rust semantics](https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast:~:text=Casting%20from%20a%20smaller,the%20source%20is%20signed) quite easy to follow. Casting from a smaller to a larger int, it just sign extends the source (smaller) int.)

.gitmodules
jayz22: I have switched back the reference.

src/next/scval_conversions.rs
```diff
@@ -166,36 +168,51 @@ pub mod int128_helpers {
...
-        u128::from(lo) | (u128::from(hi) << 64)
+    pub fn u128_from_pieces(hi: u64, lo: u64) -> u128 {
+        (u128::from(hi) << 64) | u128::from(lo)
+    }
+
+    #[must_use]
+    #[inline(always)]
+    #[allow(clippy::inline_always, clippy::cast_possible_truncation)]
+    pub fn i128_hi(i: i128) -> i64 {
+        (i >> 64) as i64
```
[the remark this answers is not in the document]
dmkozh: Actually, probably that's fine, but in any case some i128 tests with large negative numbers wouldn't hurt.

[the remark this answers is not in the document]
jayz22: Agree on the testing part. I've included a few test cases in this PR, which covers the large negative number case.
Most tests will be added in the [host PR](https://github.com/stellar/rs-soroban-env/pull/763).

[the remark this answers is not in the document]
graydon: yeah I think this is ok

graydon (approved):

PR #251, opened 2023-04-26
Remove + if dirty

sisuresh:
### What

If the submodule is dirty, then the hashes returned by `git submodule status` will be prefixed with a `+` and cause the PR to fail.


graydon (approved):

leighmcculloch:
I'm surprised that caused the PR to fail. Thanks for fixing. The intent for keeping it originally was that it would be obvious if the repo was dirty.

leighmcculloch:
Wait, why would a submodule be dirty in a PR? The submodule in the PR should always be up to date.

sisuresh:
This was actually always a problem for me, but I would just remove the `+` manually. It sounds like you don't run into this because you commit the submodule change and then run `make clean generate`?

leighmcculloch:
Ah, so this was an error you were experiencing locally, or in the PR in CI?

sisuresh:
The error would happen in CI. [This](https://github.com/stellar/rs-stellar-xdr/actions/runs/4813483107/jobs/8570061941) is what that looked like.

leighmcculloch:
That error is intended. It's important that a PR fails if the PR is not in a state that we'd want merged. If the submodule in this repo is dirty, the PR shouldn't be mergeable. 

Am I missing something about it?

sisuresh:
So are you saying my process to update xdr is incorrect? eg. I should update the submodule -> commit -> run `make clean generate` -> commit? I currently update the submodule -> run `make clean generate` -> commit.

leighmcculloch:
Ah, so I guess it doesn't matter then. You're right.

PR #255, opened 2023-05-02
Move to syn2

graydon:
Attempting to kill off dependencies on syn1


Cargo.toml
```diff
@@ -21,15 +21,15 @@ crate-git-revision = "0.0.4"
 [dependencies]
 base64 = { version = "0.13.0", optional = true }
 serde = { version = "1.0.139", features = ["derive"], optional = true }
-serde_with = { version = "2.0.1", optional = true }
+serde_with = { version = "3.0.0", optional = true }
```
leighmcculloch, line 24: We might need to test this all the way down to the CLI to check that this doesn't break some v2 feature we're dependent on. We can probably just wing this for now and risk it though as any change is likely to just be a breaking JSON format change and not something more logically sensitive.

leighmcculloch: cc @tsachiherman @paulbellamy

leighmcculloch (approved):

PR #258, opened 2023-05-14
Regenerate for overhaul of error codes

graydon:
This is the companion PR for https://github.com/stellar/stellar-xdr/pull/92 -- changing `ScStatus` to `ScError` and reassigning all error codes.


src/next/scval_conversions.rs
```diff
@@ -303,15 +303,15 @@ impl TryFrom<ScVal> for ScSymbol {
 #[cfg(feature = "alloc")]
 impl TryFrom<String> for ScVal {
     type Error = ();
-    fn try_from(v: String) -> Result<Self, Self::Error> {
+    fn try_from(v: String) -> Result<Self, ()> {
```
leighmcculloch, line 306: What's the significance of this change? I don't think there is any significance, just wondering.

src/next/scval_validations.rs
```diff
@@ -10,11 +10,11 @@ pub trait Validate {
 impl Validate for ScVal {
     type Error = Error;
 
-    fn validate(&self) -> Result<(), Self::Error> {
+    fn validate(&self) -> Result<(), <Self as Validate>::Error> {
```
leighmcculloch, line 13: Same question here.

leighmcculloch (approved):
Couple questions inline, but looks good regardless.

src/next/scval_conversions.rs
graydon: They're all just disambiguation because Self::Error is ambiguous.

src/next/scval_validations.rs
graydon: same here. any changes to returns that used Self::Error anywhere in the code are disambiguation (most of the SDK changes look like that too)

PR #274, opened 2023-06-21
Bump xdr with contract instance storage changes

dmkozh:
### What

Bump xdr with contract instance storage changes

### Why

Keeping it up to date

### Known limitations

N/A


src/next/scval_validations.rs
```diff
@@ -46,6 +45,7 @@ impl Validate for ScVal {
             }
             ScVal::Vec(None) | ScVal::Map(None) => Err(Error::Invalid),
             ScVal::Map(Some(m)) => m.validate(),
+            ScVal::Void => todo!(),
```
sisuresh, line 48: Why is this todo now?

sisuresh (approved):

src/next/scval_validations.rs
dmkozh: fixed

PR #277, opened 2023-06-23
Make git revision in version optional

leighmcculloch:
### What
Make git revision in version optional.

### Why
In some situations folks install tools in enviroments from source, but without git installed. This seems to happen to Windows users from tim eto time.

While it isn't ideal that the version isn't present, the inconvenience of users getting errors installing tools seems unreasonable to me.

A version string in the compiled product, that isn't used frequently, shouldn't be a blocker for basic use.

Note that this does open up the possibility that we break out git revision collection and end up with no revision in a production build. That's the nice thing about failing otherwise. Again, the version missing wouldn't break functionality or stop a product from working and presumably we'd fix it when we noticed.

### DIFF
--- a/src/bin/stellar-xdr/version.rs
+++ b/src/bin/stellar-xdr/version.rs
@@ -11,7 +11,10 @@ impl Cmd {
             "stellar-xdr {} ({})
 xdr (+curr): {}
 xdr (+next): {}",
-            v.pkg, v.rev, v.xdr_curr, v.xdr_next
+            v.pkg,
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Maybe I should just put this into `crate-git-revision` and have it set `"unknown"` for the `GIT_REVISION` if it's not available. 🤔 It would make builds that use it generally more resilient to environmental issues.

leighmcculloch:
Closing in favor of updating the soroban-docs saying install git.

PR #309, opened 2023-10-11
Update xdr for cost type cleanup

jayz22:
### What

### Why

### Known limitations

[TODO or N/A]


jayz22:
Didn't realize has already been updated.

PR #324, opened 2023-11-23
Render asset codes as strings in JSON

leighmcculloch:
### What
Render asset codes in JSON as UTF-8 strings, where any invalid byte sequences are encoded as a nul byte for each invalid byte.

### Why
Valid asset codes are some subset of ASCII, but they're currently rendered as hex. It creates a poor and difficult experience when a developer is expecting to see their asset code in the JSON and instead see hex.

The codes are rendered to UTF-8 strings because JSON strings must be Unicode.

Invalid byte sequences are rendered as a single nul byte per byte in the invalid sequence to preserve length. This is because the length of asset codes distinguishes between AssetCode4 and AssetCode12. Typically lossy coding of utf-8 results in Unicode Replacement Character (U+FFFD) being inserted into the output. However that character is two bytes long. If a 4-byte AssetCode4 contains an invalid byte sequence, it would result in rendering as a 5-byte UTF-8 string. When converted back to the Rust type it would be interpreted as an AssetCode12.

### DIFF
--- a/tests/str.rs
+++ b/tests/str.rs
@@ -4,8 +4,8 @@
 use stellar_xdr::curr as stellar_xdr;
 
 use stellar_xdr::{
-    AccountId, Error, Hash, MuxedAccount, MuxedAccountMed25519, NodeId, PublicKey, ScAddress,
-    SignerKey, SignerKeyEd25519SignedPayload, Uint256,
+    AccountId, AssetCode, AssetCode12, AssetCode4, Error, Hash, MuxedAccount, MuxedAccountMed25519,
+    NodeId, PublicKey, ScAddress, SignerKey, SignerKeyEd25519SignedPayload, Uint256,
 };
 
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


dmkozh (approved):

tests/str.rs
```diff
@@ -398,3 +398,146 @@ fn sc_address_from_str_with_invalid() {
...
+    assert_eq!(AssetCode::CreditAlphanum12(AssetCode12(*b"a\xd9\xaa\xd9\xaa\0\0\0\0\0\0\0")).to_string(), "a٪٪");
+    assert_eq!(AssetCode::from_str("a٪٪"), Ok(AssetCode::CreditAlphanum12(AssetCode12(*b"a\xd9\xaa\xd9\xaa\0\0\0\0\0\0\0"))));
+
+    // Round tripped to correct variant based on byte length even when utf8
+    // parsing error occurs. To preserve type consistency when round tripping
+    // the data, the length when parsing errors occur must be consistent with
+    // the input length, which is why a nul byte is expected instead of a
+    // Unicode Replacement Character, which would be two bytes.
+    assert_eq!(AssetCode::CreditAlphanum4(AssetCode4(*b"a\xc3\xc3d")).to_string(), "a\0\0d");
+    assert_eq!(AssetCode::from_str("a\0\0d"), Ok(AssetCode::CreditAlphanum4(AssetCode4(*b"a\0\0d"))));
```
leighmcculloch, line 542: Note that the side-effect of this is that usually when round tripping invalid utf8 through JSON it is detectable because you can detect the presence of the replacement character. But in this case it won't be as it'll be seen as a \0. I wonder if that is worse than having the asset type change.

dmkozh (approved):

PR #335, opened 2023-12-04
Remove scary text from readme

graydon:
it's all fairly stable and tested at this point


leighmcculloch (changes requested):
The rust docs need changing. Then use make readme to update the readme.

leighmcculloch (approved):

PR #347, opened 2024-03-04
Add schemars::JsonSchema derives to all types

leighmcculloch:
### What
Add schemars::JsonSchema derives to all types

### Why
See https://github.com/stellar/xdrgen/pull/193.

Close https://github.com/stellar/rs-stellar-xdr/issues/348

### Todo

- [x] Complete TODOs from https://github.com/stellar/xdrgen/pull/193
- [x] Add custom impls for:
   - [x] PublicKey
   - [x] AccountId
   - [x] MuxedAccount
   - [x] MuxedAccountMed25519
   - [x] SignerKey
   - [x] SignerKeyEd25519SignedPayload
   - [x] NodeId
   - [x] ScAddress
   - [x] AssetCode
   - [x] AssetCode4
   - [x] AssetCode12


tests/serde_tx_schema.rs
```diff
@@ -0,0 +1,3434 @@
...
+      }
+    },
+    "VecM_for_SorobanAuthorizedInvocation_and_4294967295": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/SorobanAuthorizedInvocation"
+      }
+    }
+  }
+}"##,
```
leighmcculloch, line 3430: Here appears a JSON schema for the Transaction Envelope. 😬 cc @willemneal

willemneal: Hmm I wonder why the numbers at the end. I'll take a closer look.

willemneal: So it seems that types like `VecM_for_SorobanAuthorizationEntry_and_4294967295` are because of how it handles generics
```rust
pub struct VecM<T, const MAX: u32 = { u32::MAX }>(Vec<T>);
```

I'll look into how we can make a special case for `VecM` to be treated like a generic array with length bounds.


```json
 "SorobanAuthorizedInvocation": {
      "description": "SorobanAuthorizedInvocation is an XDR Struct defines as:\n\n```text struct SorobanAuthorizedInvocation { SorobanAuthorizedFunction function; SorobanAuthorizedInvocation subInvocations<>; }; ```",
      "type": "object",
      "required": [
        "function",
        "sub_invocations"
      ],
      "properties": {
        "function": {
          "$ref": "#/definitions/SorobanAuthorizedFunction"
        },
        "sub_invocations": {
          "$ref": "#/definitions/VecM_for_SorobanAuthorizedInvocation_and_4294967295"
        }
      }
    },
```

Should be

leighmcculloch: Is the problem related to this issue on the schemars repo?
- https://github.com/GREsau/schemars/issues/111

Are you able to articulate the problem and post about it in that or a new issue over there?

willemneal: I'm still digging into it. I started by manually implementing `VecM`. I'm making a PR to this branch.

willemneal: Here: https://github.com/stellar/rs-stellar-xdr/pull/349

leighmcculloch: I guess we don't really need custom impls, because the issue is only with types that have generics, and the only types that do are `VecM`, `BytesM`, `StringM`, and `Frame`, which are all stock types not generated. So I should be able to copy #349 and into xdrgen.

willemneal: I suggest putting the `impl`s in a non-generated file in this crate. It will help debugging and testing.

willemneal:
@leighmcculloch I updated https://github.com/stellar/rs-stellar-xdr/pull/349

Feel free to move the impls where ever you think is best, I'm pretty happy with them now. The next issue is enum representations. See here the option is unnamed `Option 1` and the current doc strings are redundant.

Here is the site where you can view the forms generated by the schema: https://rjsf-team.github.io/react-jsonschema-form/

Unfortunately you can't share an example prefilled in, but here is a picture of the memo section of the transaction envelope. 

image

```json
  "Memo": {
    "description": "Memo is an XDR Union defines as:\n\n```text union Memo switch (MemoType type) { case MEMO_NONE: void; case MEMO_TEXT: string text<28>; case MEMO_ID: uint64 id; case MEMO_HASH: Hash hash; // the hash of what to pull from the content server case MEMO_RETURN: Hash retHash; // the hash of the tx you are rejecting }; ```",
"oneOf": [
    {
      "type": "string",
      "enum": [
        "none"
      ]
    },
    {
      "type": "object",
      "required": [
        "text"
      ],
      "properties": {
        "text": {
          "type": "string",
          "maxLength": 28
        }
      },
      "additionalProperties": false
    },
    {
      "type": "object",
      "required": [
        "id"
      ],
      "properties": {
        "id": {
          "type": "integer",
          "format": "uint64",
          "minimum": 0.0
        }
      },
      "additionalProperties": false
    },
    {
      "type": "object",
      "required": [
        "hash"
      ],
      "properties": {
        "hash": {
          "$ref": "#/definitions/Hash"
        }
      },
      "additionalProperties": false
    },
    {
      "type": "object",
      "required": [
        "return"
      ],
      "properties": {
        "return": {
          "$ref": "#/definitions/Hash"
        }
      },
      "additionalProperties": false
    }
  ]
}
```

willemneal:
Oh one other thing to note are byte arrays. Currently they are treated as `Vec<u8>`, but a base64 string would work better for JSON. Now it's pretty painful. Here is a Hash type with one input per 32 bytes.
image

leighmcculloch:
Byte arrays are rendered in the JSON as hex, not as arrays, so I'm surprised the JSON schema is rendering that as an array.

willemneal:
I update BytesM to use a hex string, but `pub struct Hash(pub [u8; 32])` is a default impl of `schemars` which is the same as `Vec<u8>` with a fixed length.

leighmcculloch:
Interesting. `Hash` does render to hex because it has the below custom serialize/deserialize based on the display and fromstr impls:
```rust
#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
#[cfg_attr(
    all(feature = "serde", feature = "alloc"),
    derive(
        serde_with::SerializeDisplay,
        serde_with::DeserializeFromStr
    )
)]
pub struct Hash(pub [u8; 32]);
```
```rust
impl core::fmt::Display for Hash {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        let v = &self.0;
        for b in v {
            write!(f, "{b:02x}")?;
        }
        Ok(())
    }
}

#[cfg(feature = "alloc")]
impl core::str::FromStr for Hash {
    type Err = Error;
    fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
        hex::decode(s).map_err(|_| Error::InvalidHex)?.try_into()
    }
}
```

willemneal:
The JsonSchema type is implemented for arrays here: https://github.com/GREsau/schemars/blob/e04e3a3a8191bac1f218539733110fcc26bdbf7c/schemars/src/json_schema_impls/array.rs#L31

Not sure how to get around other than manually implementing `Hash`?

graydon (approved):

PR #351, opened 2024-03-14
feat: add wasm-pack to allow publishing to npm

willemneal:
Currently `serde_json_wasm` caused an unreachable error. However, `serde_json` is working.

Size of current release wasm 1.6 MB.

- [ ] Add tests
- [ ] Publish to npm

### DIFF
--- a/crates/stellar-xdr-wasm/src/lib.rs
+++ b/crates/stellar-xdr-wasm/src/lib.rs
@@ -1,13 +1,23 @@
 mod utils;
 
-use std::convert::TryInto;
-
-use stellar_xdr::schema::Schema;
+use stellar_xdr::{curr::TypeVariant, schema::Schema};
 use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
-pub fn schema() -> String {
-    let Schema(json_schema): Schema = stellar_xdr::curr::TypeVariant::Transaction.try_into().unwrap();
+pub fn type_variants() -> String {
+    format!(
+        "[{}]",
+        TypeVariant::VARIANTS_STR
+            .iter()
+            .map(|s| format!("{s:?}"))
+            .collect::<Vec<String>>()
+            .join(", ")
+    )
+}
+
+#[wasm_bindgen]
+pub fn schema(s: &str) -> String {
+    let Schema(json_schema): Schema = s.parse().unwrap();
     let s = match serde_json::to_string(&json_schema) {
         Ok(s) => s,
         Err(e) => return format!("{{\"error\": \"{}\"}}", e),
--- a/crates/stellar-xdr-wasm/src/lib.rs
+++ b/crates/stellar-xdr-wasm/src/lib.rs
@@ -1,0 +1,16 @@
+mod utils;
+
+use std::convert::TryInto;
+
+use stellar_xdr::schema::Schema;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub fn schema() -> String {
+    let Schema(json_schema): Schema = stellar_xdr::curr::TypeVariant::Transaction.try_into().unwrap();
+    let s = match serde_json::to_string(&json_schema) {
+        Ok(s) => s,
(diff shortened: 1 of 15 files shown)


src/schema.rs
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                } else if let Some(enum_values) = &obj.enum_values {
+                    if let Some(title) = enum_values.get(2) {
+                        obj.metadata().title = Some(title.to_string());
+                    }
+                }
+            }
+        });
+    }
+}
```
leighmcculloch, line 68: Nice. We should move this into the generated schema so that no matter how the schema is being generated this same logic is in use. I'll have a look at copying this into:
- https://github.com/stellar/xdrgen/pull/193

src/schema.rs
```diff
@@ -0,0 +1,68 @@
...
+impl FromStr for Schema {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Schema, Error> {
+        s.parse::<crate::curr::TypeVariant>()
+            .map_err(|_| {
+                Error::UnknownType(s.to_string(), &crate::curr::TypeVariant::VARIANTS_STR)
+            })?
+            .try_into()
+    }
+}
```
leighmcculloch, line 36: What's depending on the TryFrom and the FromStr?

I'm not sure how we can support different versions of schema with them, and it looks like they primarily exist to inject the ReplaceAdditionalProperties which we should really do at the generation of the schema in xdrgen.

We'd also need these to be embedded under curr and next.

crates/stellar-xdr-wasm/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    serde_json_wasm::to_string(&json_schema).map_err(|e| JsValue::from_str(&format!("{e:?}")))
+}
+
+#[wasm_bindgen]
+pub fn from_xdr(xdr_base64: String, variant: Option<String>) -> String {
+    let mut f = curr::Limited::new(ResetRead::new(xdr_base64.as_bytes()), curr::Limits::none());
+    for variant in variant
+        .map(|v| Some(vec![v.parse::<crate::curr::TypeVariant>().ok()?]))
+        .unwrap_or_else(|| Some(TypeVariant::VARIANTS.to_vec()))
+        .unwrap()
```
leighmcculloch, line 42: I think the from_xdr function, by providing an auto-search for a value, is providing higher level functionality at a lower level than we need to. If we instead expose a `pub fn types() -> Vec<TypeVariant>`, then any caller will be able to iterate over the variants and try them out. Some advantages: keep the logic in the lib simpler, no need to maintain and document string values that have a special form `{"error":...}` that could actually collide with future valid XDR json.

crates/stellar-xdr-wasm/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[wasm_bindgen]
+pub fn type_variants() -> String {
+    format!(
+        "[{}]",
+        TypeVariant::VARIANTS_STR
+            .iter()
+            .map(|s| format!("{s:?}"))
+            .collect::<Vec<String>>()
+            .join(", ")
```
leighmcculloch, line 23: We should remove the space so that the list is comma separated values that don't need to be trimmed. A UI / presentation layer can add in any spaces between values if it is presenting the values that way.
```suggestion
            .join(",")
```

leighmcculloch, line 23: We should be able to return a `JsValue` and use the `Array` type to build an array of strings to return to the JS side. Is that compatible with what we're doing here?

leighmcculloch (commented):
👏🏻 Some comments inline, I think we'll want to move some of the logic once this is in a working state and we've worked out any issues with binary size or embedding into the lab.

leighmcculloch:
I'm picking this change up, but am continuing in https://github.com/stellar/rs-stellar-xdr/pull/360 because for some reason even though the checkbox is showing a green tick next to "maintainers are allowed to edit this pull request," I seem to get a 403 error whenever I try to push to the branch.

PR #360, opened 2024-04-10
Add JavaScript bindings via Wasm

leighmcculloch:
### What
Add JavaScript bindings via Wasm

### Why
Based on #347 and continuation of #351 by @willemneal.

The Rust XDR lib is the only implementation of the XDR<>JSON format. This is a shortcut to allowing the use of the format in JavaScript contexts as a hold over until there is a native implementation in JavaScript.

Everything in this change is experimental, even more experimental than the JSON format that is itself still relatively experimental.

### Merging
Intended to be merged to `main` after #347.


leighmcculloch:
Superseded by:
- https://github.com/stellar/js-stellar-xdr-json/pull/1

PR #368, opened 2024-05-13
Encode claimable balance ids to match Horizon in JSON

### LINKED ISSUE #367: Render claimable balance IDs in the same format that Horizon renders them in
Claimable balances are rendered as a hex string contained in an object with the v0 discriminant.

In Horizon, claimable balance IDs are rendered by hex encoding the XDR for the object structure.

Given that folks are used to the Horizon rendering, we should use that in the JSON <> XDR as well.

```diff
           "source_account": null,
           "body": {
             "claim_claimable_balance": {
-              "balance_id": {
-                "claimable_balance_id_type_v0": "d0452b955a0c05c4465ebc9192a23db50ce6aa719b4b61c465bb2675e83b0a05"
-              }
+              "balance_id": "00000000d0452b955a0c05c4465ebc9192a23db50ce6aa719b4b61c465bb2675e83b0a05"
             }
           }
         }
```

Ref: https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1715267848560049

Note: There's no SEP or specification that captures the claimable balance ID format.

leighmcculloch:
### What
Encode claimable balance ids to match Horizon in JSON

### Why
Claimable balances are rendered as a hex string contained in an object with the v0 discriminant.

In Horizon, claimable balance IDs are rendered by hex encoding the XDR for the object structure.

Given that folks are used to the Horizon rendering, we should use that in the JSON <> XDR as well.

Close #367

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ CARGO_HACK_ARGS=--feature-powerset --exclude-features default --group-features b
 CARGO_DOC_ARGS?=--open
 
 XDRGEN_VERSION=e2cac557162d99b12ae73b846cf3d5bfe16636de
-XDRGEN_TYPES_CUSTOM_STR_IMPL_CURR=PublicKey,AccountId,MuxedAccount,MuxedAccountMed25519,SignerKey,SignerKeyEd25519SignedPayload,NodeId,ScAddress,AssetCode,AssetCode4,AssetCode12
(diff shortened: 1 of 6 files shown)


leighmcculloch:
I'd like to merge this change to accompany https://github.com/stellar/rs-stellar-xdr/pull/347 and https://github.com/stellar/rs-stellar-xdr/pull/360, and release them all in a minor release. However, since this change changes the JSON format, it is a breaking change.

We've always said internally that the JSON format was experimental and unstable, but never actually said that in the docs for the crate. In any case it's pretty important not to introduce a breaking change into the JSON anywhere it'd cause the CLI to break with data it caches to disk.

Given claimable balances don't overlap with Soroban usage at all and the only CLI usage is for Soroban only txs, it's probably an extremely low likelihood and low impact change to make.

cc @accordeiro @dmkozh @quietbits

dmkozh (approved):

PR #370, opened 2024-05-14
feat: add From<Transaction> and From<&Transaction>

willemneal:
### What

Tools need to always expect TransactionEvleope at the headings. This makes it easy to go in one direction.

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/src/curr/mod.rs
+++ b/src/curr/mod.rs
@@ -6,6 +6,7 @@ mod str;
 
 mod scval_conversions;
 pub use scval_conversions::*;
+mod transaction_conversions;
 
 mod scval_validations;
 pub use scval_validations::*;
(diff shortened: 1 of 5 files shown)


willemneal:
@leighmcculloch should we also include a `TryFrom<TransactionEnvelope>`?  If so would we need upstream changes to add an error case?

leighmcculloch (commented):
I left a few comments inline.

I think we can add this conversion because it is common to build transactions and to want an envelope. It's also reasonable to do the same for FeeBumpTransaction too, so I think that could be added.

> should we also include a TryFrom<TransactionEnvelope>? If so would we need upstream changes to add an error case?

I don't think we should add reverse conversions. The reverse conversion would discard data which could result in bugs for round trip operations, which are often the types of operations we perform with transactions. And the `match` keyword is convenient enough and ensures devs handle all cases or be very explicitly not handling some.

For the existing tests, I'd leave them building TEs, as there's little succinctness introduced at the cost of the .into() obfuscating.

Because the comments inline are small, I'll update them and merge the change.

willemneal:
@leighmcculloch I removed the edits to the previous tests.

leighmcculloch (approved):

PR #384, opened 2024-08-14
Bump Rust version to 1.79

dmkozh:
### What

Bump Rust version to 1.79

### Why

We're doing this bump for env and we need to fix the XDR library warnings anyways.

### Known limitations

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Stellar Development Foundation <info@stellar.org>"]
 license = "Apache-2.0"
 version = "21.2.0"
 edition = "2021"
-rust-version = "1.74.0"
+rust-version = "1.79.0"
 
 [[bin]]
 name = "stellar-xdr"
@@ -15,7 +15,7 @@ path = "src/bin/stellar-xdr/main.rs"
 required-features = ["cli"]
 doctest = false
 
-[build_dependencies]
+[build-dependencies]
 crate-git-revision = "0.0.6"
 
 [dependencies]
@@ -31,7 +31,7 @@ serde_json = { version = "1.0.89", optional = true }
 thiserror = { version = "1.0.37", optional = true }
 schemars = { version = "0.8.16", optional = true }
 
-[dev_dependencies]
+[dev-dependencies]
 serde_json = "1.0.89"
 
 [features]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -7,15 +7,15 @@ authors = ["Stellar Development Foundation <info@stellar.org>"]
 license = "Apache-2.0"
 version = "21.2.0"
 edition = "2021"
-rust-version = "1.74.0"
+rust-version = "1.79.0"
```
leighmcculloch, line 10: This value doesn't set the rust-version we build with, it sets the minimum rust version that importers can use, which means it forces folks using the lib to update to rust 1.79. CI builds with both the min version and the latest version:
https://github.com/stellar/rs-stellar-xdr/blob/0eebf75944aee8c25578dc0fd4a2ad6e1a4867ee/.github/workflows/rust.yml#L74-L85

We typically only need to update this value if we're planning to use features in a newer release.

What's the reason for selecting 1.79.0 as the minimum rust version? It's not the latest version.

leighmcculloch (commented):
We can update the minimum supported rust version, but it's not clear we need to, because the PR description isn't clear as to why the version is being updated. No changes in the PR are using new rust features that require a newer version.

Cargo.toml
dmkozh: >  because the PR description isn't clear as to why the version is being updated. 

We're bumping env to 1.79 and are running into warnings in XDR library that are treated like errors. Bumping the version would make sure that we don't accidentally introduce new warnings.

>  it sets the minimum rust version that importers can use, which means it forces folks using the lib to update to rust 1.79.

Yeah, but the env update will force the most users to update anyway - not sure update is inherently problematic.

> What's the reason for selecting 1.79.0 as the minimum rust version? It's not the latest version.

Not sure, that's what Graydon has used in the env PR - maybe it's the minimum version that Wasmi supports.

In any case, I just want to unblock the env bump, I can leave the version not bumped.

sisuresh (approved):

Cargo.toml
dmkozh: I'll only merge the non-controversial warning fixes for now.

PR #387, opened 2024-08-28
Support encoding streams in the cli

leighmcculloch:
### What
Support encoding streams in the cli.

### Why
So that it is possible to pipe streams of XDR values to the CLI encode command and get out a stream of encoded XDR. Useful for building valid contract spec entries for WASMs using JSON.

You might also notice that stream base64 format is not supported for outputting. It would require more significant refactoring of this area of code to support that, and nothing in this PR does anything new to preclude doing that in the future, so I'm deferring it since I don't need it in this moment.

Dependent on:
- https://github.com/stellar/xdrgen/pull/203

### DIFF
--- a/src/cli/encode.rs
+++ b/src/cli/encode.rs
@@ -138,7 +138,7 @@ macro_rules! run_x {
                                     Err(e) => Err(e)?,
                                 };
                                 let l = crate::$m::Limits::none();
-                                stdout().write_all(&t.to_xdr(l.clone())?)?
+                                stdout().write_all(&t.to_xdr(l)?)?
                             }
                         }
                     },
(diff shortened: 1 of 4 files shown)


src/cli/encode.rs
```diff
@@ -96,6 +96,9 @@ impl Default for InputFormat {
 pub enum OutputFormat {
     Single,
     SingleBase64,
+    Stream,
+    // TODO: StreamBase64,
```
sisuresh, line 100: Do we need to open issues for these?

sisuresh (approved):

src/cli/encode.rs
leighmcculloch: We could, but there is no pressing need for these things. So I think leaving the TODOs here as a way to signal if that feature is added here's the area to add it is okay. But I can also remove the TODOs too.

leighmcculloch: Opened issue:
- https://github.com/stellar/rs-stellar-xdr/issues/388

PR #391, opened 2024-09-03
Regen to pick up bls12381 cost types

jayz22:
### What

### Why

### Known limitations

[TODO or N/A]


jayz22:
Did notice it's already done https://github.com/stellar/rs-stellar-xdr/pull/390

PR #396, opened 2024-09-25
Add conversions from AccountId and PublicKey to MuxedAccount

leighmcculloch:
### What
Add conversions from AccountId and PublicKey to MuxedAccount.

### Why
The types are largely compatible and while MuxedAccounts are used in transactions applications often only support G addresses and convenient conversions from the AccountId/PublicKey type used for G addresses to the MuxedAccount type for use in transactions would be helpful.


leighmcculloch:
@willemneal What other conversions would be helpful iyo?

Issue #406, opened 2025-01-13
cli: accept xdr via an argument and not only as a file or stdin


leighmcculloch:
The stellar-xdr cli (which is also the stellar xdr command in the stellar-cli) accepts as input xdr either via file or stdin.

e.g. via file:

```bash
stellar xdr decode myfile.xdr myfile2.xdr
```

e.g. via stdin:

```bash
cat myfile.xdr | stellar xdr decode
```

```bash
echo ... | stellar xdr decode
```

It's been reported that it would be convenient if the command provided a way to pass the xdr as an argument as well.

When the command was first implemented passing the xdr as an argument was omitted because in bash-like environments you get that functionality for free.

e.g.

```bash
stellar xdr decode <(echo ...)
```

However, the stellar-cli is supported across a variety of environments, including Windows, where bash may not be in use. For the sake of making the cli consistently usable on all environments, which simplifies tutorials and teaching material (e.g. conference slides), we could add the ability to pass the xdr as an argument.

We should only do this if it doesn't make the rest of the UX more complex, and only if this is high impact enough.

cc @stellar/devx @tomerweller

leighmcculloch:
The first thought I had is this:

```bash
# existing behaviour preserved
stellar xdr decode myfile.xdr myfile2.xdr # assumes file because files passed
cat myfile.xdr | stellar xdr decode # assumes stdin because no files given
echo ... | stellar xdr decode # assumes stdin because no files given

# new behavior added
stellar xdr decode --read files myfile.xdr myfile2.xdr # no assumption because parameter specified
cat myfile.xdr | stellar xdr decode --read stdin # no assumption because parameter specified
echo ... | stellar xdr decode --read stdin # assumes stdin because no files given
stellar xdr decode --read args AAAAAAjBALVT2rv9LPtsuKN2lrdTALY5naljcz5sujEVXHX6
```

But that's quite complex. Easy enough to read, but difficult to discover.

PR #407, opened 2025-01-22
Add support decoding/encoding strings/files as arguments

Ifropc:
### What
Adds support to decode/encode strings as arguments, e.g.

```
> cargo run --features=cli decode --type TransactionEnvelope `cat /tmp/test` /tmp/test

{"tx":{"tx":{"source_account":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","fee":100,"seq_num":123,"cond":{"time":{"min_time":0,"max_time":0}},"memo":"none","operations":[{"source_account":null,"body":{"create_account":{"destination":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","starting_balance":1230000000}}}],"ext":"v0"},"signatures":[]}}
{"tx":{"tx":{"source_account":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","fee":100,"seq_num":123,"cond":{"time":{"min_time":0,"max_time":0}},"memo":"none","operations":[{"source_account":null,"body":{"create_account":{"destination":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","starting_balance":1230000000}}}],"ext":"v0"},"signatures":[]}}

```

```
cargo run --features=cli encode --type TransactionEnvelope `cat /tmp/transaction` /tmp/transaction

AAAAAgAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAGQAAAAAAAAAewAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAABJUE+AAAAAAAAAAAA=
AAAAAgAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAGQAAAAAAAAAewAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAABJUE+AAAAAAAAAAAA=
```

Files that was used:
```
cat /tmp/test
AAAAAgAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAGQAAAAAAAAAewAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAC9yj33uT0b+EZcua+UlGswq2VGc3P8HY5pYp1d3N/YWwAAAABJUE+AAAAAAAAAAAA=

cat /tmp/transaction
{"tx":{"tx":{"source_account":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","fee":100,"seq_num":123,"cond":{"time":{"min_time":0,"max_time":0}},"memo":"none","operations":[{"source_account":null,"body":{"create_account":{"destination":"GC64UPPXXE6RX6CGLS427FEUNMYKWZKGONZ7YHMONFRJ2XO437MFXFGO","starting_balance":1230000000}}}],"ext":"v0"},"signatures":[]}}
```

### Why

#406 

### Known limitations

N/A


.gitignore
```diff
@@ -1 +1,2 @@
 /target
+.idea
```
leighmcculloch, line 2: Can you put this into your personal global gitignore file? See the following blog for instructions for how to set it up:
https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

src/cli/decode.rs
```diff
@@ -77,44 +76,52 @@ impl Default for OutputFormat {
...
-                match self.input {
-                    InputFormat::Single => {
-                        let mut l = crate::$m::Limited::new(f, crate::$m::Limits::none());
-                        let t = crate::$m::Type::read_xdr_to_end(r#type, &mut l)?;
-                        self.out(&t)?;
-                    }
-                    InputFormat::SingleBase64 => {
+
+            match self.input_format {
+                InputFormat::Single => consume_input_or_err(&self.input, |f: Box<dyn Read>| {
```
leighmcculloch, line 84: Why did the arrangement of this code change from getting a list of `Box<dyn Read>` and then calling the code on each item in the list, to calling a function that then does a callback for each file?

I would have thought this change could be much more surgical, simply expanding the input possibilities and those new inputs wouldn't result in any change to this area of code because string inputs would just be another Box<dyn Read>.

leighmcculloch (commented):
Question inline because there's more changing in this PR than I would have expected given the goal.

src/cli/decode.rs
Ifropc: I originally wanted to do it like that, so we simply feed bytes to the transformer (decoder/encoder) function, but `consume_input` function has logic that if file exists, it will apply transformer logic both to file and to the string, and will return an error if both are valid. 
I'm not sure if that's the simplest approach (probably not, but I couldn't figure out the way of doing it easier).
As a bonus point, consume_input can be used from cli and it already has all the logic in it (including consuming stdin). 
In the future we can expand it (e.g. to automatically recognize XDR base64/json/etc) and it will go in all our tooling without changing outside logic.

leighmcculloch: Below is what the change would look like. The structure and much of the app wouldn't need to change, the process of loading the file just needs to handle the error case if the file is not found.

Also note the use of `OsString`, because file paths are not always compatible with `String`.

```diff
diff --git a/src/cli/decode.rs b/src/cli/decode.rs
index 5521249..ad75acf 100644
--- a/src/cli/decode.rs
+++ b/src/cli/decode.rs
@@ -1,8 +1,8 @@
 use std::{
+    ffi::OsString,
     fmt::Debug,
     fs::File,
-    io::{stdin, Read},
-    path::PathBuf,
+    io::{stdin, Cursor, ErrorKind, Read},
     str::FromStr,
 };
 
@@ -28,9 +28,9 @@ pub enum Error {
 #[derive(Args, Debug, Clone)]
 #[command()]
 pub struct Cmd {
-    /// Files to decode, or stdin if omitted
+    /// Files to decode, or XDR base64 strings, or stdin if omitted
     #[arg()]
-    pub files: Vec<PathBuf>,
+    pub files: Vec<OsString>,
 
     /// XDR type to decode
     #[arg(long)]
@@ -141,14 +141,17 @@ impl Cmd {
         if self.files.is_empty() {
             Ok(vec![Box::new(stdin())])
         } else {
-            Ok(self
-                .files
+            self.files
                 .iter()
-                .map(File::open)
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .map(|f| -> Box<dyn Read> { Box::new(f) })
-                .collect())
+                .map(|input| match File::open(input) {
+                    Ok(file) => Ok(Box::new(file) as Box<dyn Read>),
+                    Err(e) if e.kind() == ErrorKind::NotFound => {
+                        Ok(Box::new(Cursor::new(input.clone().into_encoded_bytes()))
+                            as Box<dyn Read>)
+                    }
+                    Err(e) => Err(e.into()),
+                })
+                .collect()
         }
     }
 
diff --git a/src/cli/encode.rs b/src/cli/encode.rs
index b8892b9..4bd64ec 100644
--- a/src/cli/encode.rs
+++ b/src/cli/encode.rs
@@ -1,7 +1,7 @@
 use std::{
+    ffi::OsString,
     fs::File,
-    io::{stdin, stdout, Read, Write},
-    path::PathBuf,
+    io::{stdin, stdout, Cursor, ErrorKind, Read, Write},
     str::FromStr,
 };
 
@@ -64,9 +64,9 @@ impl From<crate::next::Error> for Error {
 #[derive(Args, Debug, Clone)]
 #[command()]
 pub struct Cmd {
-    /// Files to encode, or stdin if omitted
+    /// Files to encode, or JSON strings, or stdin if omitted
     #[arg()]
-    pub files: Vec<PathBuf>,
+    pub files: Vec<OsString>,
 
     /// XDR type to encode
     #[arg(long)]
@@ -172,14 +172,17 @@ impl Cmd {
         if self.files.is_empty() {
             Ok(vec![Box::new(stdin())])
         } else {
-            Ok(self
-                .files
+            self.files
                 .iter()
-                .map(File::open)
-                .collect::<Result<Vec<_>, _>>()?
-                .into_iter()
-                .map(|f| -> Box<dyn Read> { Box::new(f) })
-                .collect())
+                .map(|input| match File::open(input) {
+                    Ok(file) => Ok(Box::new(file) as Box<dyn Read>),
+                    Err(e) if e.kind() == ErrorKind::NotFound => {
+                        Ok(Box::new(Cursor::new(input.clone().into_encoded_bytes()))
+                            as Box<dyn Read>)
+                    }
+                    Err(e) => Err(e.into()),
+                })
+                .collect()
         }
     }
 }

```

Ifropc: I was thinking about something like that, but it wouldn't handle ambiguous strings (both a valid file path and an xdr), so I didn't go with it originally, do you think we should just not implement it for now?

leighmcculloch: If the input matches a file, should probably go with it being a file, as it's very unlikely base64 xdr would be a filename.

Ifropc: I tried going with this suggestion and another issue that I noticed is we will try to read file with invalid filename (e.g. json string) that can be too long, contain invalid characters, etc. 
The error for that is [unstable](https://github.com/rust-lang/rust/blob/master/library/std/src/io/error.rs#L377-L378) and I believe it will differ depending on the system. 
I think the options are either handle filename errors (sounds bug-prone and we would need to test it across all platforms), or try to parse string on any file error (will give bad UX as we ignore file errors) 
WDYT?

leighmcculloch: We can just include invalid file name as a case that falls back to xdr decode?

We can also try decode xdr then read file, my main point is the change can be much more surgical, we only need to change the files function to insert the new logic.

leighmcculloch: > The error for that is [unstable](https://github.com/rust-lang/rust/blob/master/library/std/src/io/error.rs#L377-L378)

Ah sorry, I missed this. Yeah checking the file exists first makes sense.

Ifropc: Sorry just trying to understand what should we do with filename error (we can't check for it because error type is unstable and we don't use nightly)?

leighmcculloch: Yup, so do the file.exists() check you had before rather than relying on the error. But the change as a whole should be a lot smaller now yeah?

leighmcculloch (approved):
👏🏻 Looks great. Fyi the `guess` command also needs updating.

PR #409, opened 2025-02-01
feat(CLI): make schema command public

willemneal:
This is needed for allowing the CLI to write the schema for VScode and other editors which support json schema.

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/src/cli/types.rs
+++ b/src/cli/types.rs
@@ -1,5 +1,5 @@
 pub mod list;
-mod schema;
+pub mod schema;
 
 use clap::{Args, Subcommand};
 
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Unsure if I'd build on the cli mods. But happy to mark it pub. They are only exposed publicly for the stellar-cli to reexport. So warning they may break in service of the cli.

leighmcculloch:
Looks like a couple docs need adding for the build to pass. Because the mod is pub now. 

If you fix that I can merge it.

willemneal:
@leighmcculloch It should be good now!

leighmcculloch (approved):

PR #412, opened 2025-02-20
Add support of strings/files as arguments to guess + refactoring

Ifropc:
### What

Followup for #407 

* Adds support of guessing string and files 
* Refactored original implementation to remove copy-paste and slightly improve UX of guess command
 
### Why

Part of #406 

### Known limitations

N/A

### DIFF
--- a/src/cli/guess.rs
+++ b/src/cli/guess.rs
@@ -1,11 +1,10 @@
-use clap::{Args, ValueEnum};
 use std::ffi::OsString;
 use std::{
     cmp,
-    fs::File,
-    io::{self, stdin, Read},
-    path::PathBuf,
+    io::{self, Read},
 };
+
+use clap::{Args, ValueEnum};
 
 use crate::cli::{skip_whitespace::SkipWhitespace, util, Channel};
 
--- a/src/cli/guess.rs
+++ b/src/cli/guess.rs
@@ -1,3 +1,5 @@
+use clap::{Args, ValueEnum};
+use std::ffi::OsString;
 use std::{
     cmp,
     fs::File,
@@ -5,9 +7,7 @@ use std::{
     path::PathBuf,
 };
 
-use clap::{Args, ValueEnum};
-
-use crate::cli::{skip_whitespace::SkipWhitespace, Channel};
+use crate::cli::{skip_whitespace::SkipWhitespace, util, Channel};
 
 #[derive(thiserror::Error, Debug)]
 #[allow(clippy::enum_variant_names)]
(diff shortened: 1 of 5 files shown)


src/cli/guess.rs
```diff
@@ -69,68 +72,80 @@ impl Default for OutputFormat {
...
-                        for v in iter {
-                            match v {
-                                Ok(_) => count += 1,
-                                Err(_) => continue 'variants,
+            let mut input = util::parse_input::<Error>(&self.input)?;
+            let len = input.len();
+            for (i, read) in &mut input.iter_mut().enumerate() {
+                let mut rr = ResetRead::new(read);
+                let mut guessed = false;
+                'variants: for v in crate::$m::TypeVariant::VARIANTS {
```
Ifropc, line 80: FYI: no changes in actual implementation

src/cli/guess.rs
```diff
@@ -18,22 +17,24 @@ pub enum Error {
...
 }
 
 #[derive(Args, Debug, Clone)]
 #[command()]
 pub struct Cmd {
-    /// File to decode, or stdin if omitted
+    /// XDR or files containing XDR to decode, or stdin if empty
     #[arg()]
-    pub file: Option<PathBuf>,
+    pub input: Vec<OsString>,
```
leighmcculloch, line 29: Afaict there is a change in behaviour here as part of this refactor. The input was changed from being one path, to be a list of paths or inputs. This command's API is intended to return only a list of possible types for that one value. If it returns a list of possible types for multiple inputs, the output is going to be pretty confusing.

Unclear on the motivation of the refactor here, the guess command doesn't need to use parse_input, it's `file` function is already very simple. We don't need the code to be super DRY as it isn't material debt.

src/cli/mod.rs
```diff
@@ -65,7 +66,8 @@ impl Default for Channel {
 pub enum Cmd {
     /// View information about types
     Types(types::Cmd),
-    /// Guess the XDR type
+    /// Guesses the XDR type. All possible types are be printed for given XDR using given output format
+    /// (list views with each item on a new line by default).
```
leighmcculloch, line 70: I suggest we format this slightly different.
```suggestion
    /// Guesses the XDR type.
    ///
    /// Prints a list of types that the XDR values can be decoded into.
```

src/cli/guess.rs
```diff
@@ -131,8 +135,12 @@ macro_rules! run_x {
                 };
                 if count > 0 {
                     println!("{}", v.name());
+                    guessed = true;
                 }
             }
+            if (!guessed) {
+                return Err(Error::InvalidType);
+            }
```
leighmcculloch, line 143: This looks like a change that wasn't planned or discussed, and is unrelated to the change that's required for this issue. Could we move it into a separate pull request to discuss separately? Because the command returns a list that can be iterated on, returning an error in this case may make the command less composable. We might want to only return an error if `--exit-code` is passed.

The error would need renaming. Invalid type makes it sound like the user has specified a type. It may actually map to a valid type, but maybe a different input format needs selecting.

Ifropc: Okay, let's do that!

Ifropc: Created a task for that #413

src/cli/mod.rs
```diff
@@ -65,7 +66,9 @@ impl Default for Channel {
 pub enum Cmd {
     /// View information about types
     Types(types::Cmd),
-    /// Guess the XDR type
+    /// Guesses the XDR type.
```
leighmcculloch, line 69:
```suggestion
    /// Guess the XDR type.
```

Ifropc:
@leighmcculloch just FYI: this PR can't be merged without approval :)

leighmcculloch (approved):

PR #418, opened 2025-03-08
Add generated xdr-json schema files

leighmcculloch:
### What

Add generated xdr-json schema files.

### Why

So that XDR-JSON schema files can be referenced via a URL for any commit / build / release of the stellar-xdr crate.

So that tooling that embeds the stellar-xdr crate (CLI, Lab, RPC), and supports parsing XDR-JSON and converting from XDR to and from JSON, can also reference a schema URL for IDEs and editors like VS Code that can use JSON schema to auto-complete and provide hints about valid and invalid JSON.

The generated JSON files are checked by CI via the existing `generate` job, that doesn't need updating to do the checks because it broadly checks for any diff after regeneration.

As a bonus, after this change is merged, any future changes to the XDR will be more clearly evident as to how they impact the XDR-JSON schema.

### Known limitations

This change doesn't yet embed a `$schema` key into JSON generated or parsed yet. It is unclear yet if we will add that here, or not, because not all uses of the XDR-JSON would benefit from it. So applications, like the CLI, that will use it, may do so as needed separately.

This change doesn't result in the JSON being served on a particularly friendly URL, but we will be able to use the stellar.org nginx's ability to redirect paths to redirect a nicer stellar.org URL to the files hosted here.

There's an argument for hosting these files outside this repo, however hosting them here is a trade-off towards simplifying the moving parts that support this.

cc @stellar/devx @fnando


dmkozh (approved):

fnando (approved):
nice!

ps: it's funny how this pr made github unusable. 🫠

leighmcculloch:
@fnando I was debating limiting the files written to the TransactionEnvelope, at least for the moment, so as not to significantly burden the file system with the +457,000 lines. I'm hoping the GH slowness is just because of the size of the PR, and will not continue once merged.

PR #419, opened 2025-03-14
Regenerate XDR to pick up CAP-67 changes.

dmkozh:
### What

Regenerate XDR to pick up CAP-67 changes.

### Why

Keeping it up to date

### Known limitations

N/A


src/next/str.rs
```diff
@@ -71,6 +72,19 @@ impl core::str::FromStr for AccountId {
...
+        self.0.fmt(f)
+    }
+}
+
+impl core::str::FromStr for PoolId {
+    type Err = Error;
+    fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
+        Ok(PoolId(Hash::from_str(s)?))
+    }
+}
```
leighmcculloch, line 86: These needs to use the strkey library change here:
- https://github.com/stellar/rs-stellar-strkey/pull/74

I was waiting on a review of the SEP-23 change here by @stanford-scs:
- https://github.com/stellar/stellar-protocol/pull/1646

src/curr/str.rs
```diff
@@ -253,14 +272,25 @@ impl core::str::FromStr for ScAddress {
...
             ScAddress::Contract(Hash(h)) => {
                 let k = stellar_strkey::Contract(*h);
                 let s = k.to_string();
-                f.write_str(&s)?;
+                f.write_str(&s)
+            }
+            ScAddress::MuxedAccount(muxed_ed25519_account) => {
+                let k = stellar_strkey::Strkey::MuxedAccountEd25519(
+                    stellar_strkey::ed25519::MuxedAccount {
+                        ed25519: muxed_ed25519_account.ed25519.0.clone(),
```
leighmcculloch, line 284: Clone shouldn't be necessary here I think because the type should be an array?

```suggestion
                        ed25519: muxed_ed25519_account.ed25519.0,
```

src/curr/str.rs
```diff
@@ -253,14 +272,25 @@ impl core::str::FromStr for ScAddress {
...
+                        id: muxed_ed25519_account.id,
+                    },
+                );
+                let s = k.to_string();
+                f.write_str(&s)
             }
+            ScAddress::ClaimableBalance(claimable_balance_id) => claimable_balance_id.fmt(f),
+            ScAddress::LiquidityPool(pool_id) => pool_id.fmt(f),
```
leighmcculloch, line 292: Same as https://github.com/stellar/rs-stellar-xdr/pull/419/files#r1994488116, can you drop some TODOs here to fix this up to use the strkey library once available?

src/next/str.rs
```diff
@@ -21,12 +21,13 @@
...
 //# ## Other
 //# - ClaimableBalanceId
+//# - PoolId
 #![cfg(feature = "alloc")]
 
 use super::{
     AccountId, AssetCode, AssetCode12, AssetCode4, ClaimableBalanceId, Error, Hash, MuxedAccount,
-    MuxedAccountMed25519, NodeId, PublicKey, ScAddress, SignerKey, SignerKeyEd25519SignedPayload,
-    Uint256,
+    MuxedAccountMed25519, MuxedEd25519Account, NodeId, PoolId, PublicKey, ScAddress, SignerKey,
```
leighmcculloch, line 29: Can you add a Display and FromStr impl for the `MuxedEd25519Account` type? You'll need to include the type in the `Makefile`s list of types to exclude the derived variety for. Same for PoolId and ClaimableBalanceId when they're implemented.

leighmcculloch (commented):
Couple small asks, otherwise looks good.

src/next/str.rs
dmkozh: Yeah, I've opened an issue https://github.com/stellar/rs-stellar-xdr/issues/420 and added a TODO.

src/curr/str.rs
dmkozh: Done.

src/curr/str.rs
dmkozh: Done

src/next/str.rs
dmkozh: Done. I've also updated the makefile, though I'm not sure what exactly this does.

graydon (approved):

PR #421, opened 2025-03-17
Add LedgerEntry to LedgerKey conversion

leighmcculloch:
### What
Add to_key() conversion method for LedgerEntry and it's sub-types types.

### Why
Enable direct conversion from LedgerEntry to LedgerKey to simplify working with ledger entries. In some application code we've had to construct ledger keys from ledger entries, and having this conversion logic in the stellar-xdr lib would be helpful.


src/next/ledgerkey.rs
```diff
@@ -0,0 +1,220 @@
...
+        LedgerKeyContractCode {
+            hash: self.hash.clone(),
+        }
+    }
+}
+
+impl ConfigSettingEntry {
+    #[must_use]
+    pub fn to_key(&self) -> LedgerKeyConfigSetting {
+        let config_setting_id = match self {
```
dmkozh, line 123: Why not `self.discriminant()`?

leighmcculloch: Oh nice. Good call. 9ffe57e074529cb4d4e515483675c869e26bcd8f

dmkozh (approved):

PR #424, opened 2025-03-19
Add arbitrary and default commands to CLI

leighmcculloch:
### What
Add generate command for generating arbitrary XDR values and default XDR values.

### Why
For quick generation of arbitrary and default values for xdr types at runtime given an enum. Will be used in testing and some UI components. For example, getting arbitrary structure can be helpful in testing. And getting a default value can be used in a text editor to provide a default shape of an XDR structure as a starting point.

The change ensures a default impl is available for every type, but allows custom default impls to be written by the rs-stellar-xdr crate for those that would make more sense to have some non-default default value. An example would be the TransactionEnvelope union that an automated default would result in a v0 transaction instead of the more common v1.

### Merging

Dependent on:
- https://github.com/stellar/xdrgen/pull/216


src/cli/generate/default.rs
leighmcculloch, line 1: I don't love the name 'default' for this command. 'template' might be a better name. I didn't want to dwell on it though so left it as default, that matches the Rust behavior.

dmkozh (approved):

PR #450, opened 2025-05-16
Fix Error trait impl

leighmcculloch:
### What
  Fix the trait impls of the Error type that had gaps, missing variant comparisons and missing impls in getting source errors.

  ### Why
  The current implementation incorrectly compares errors and is missing some impls on some error variants, such as being able to get the source.

Supersedes:
- https://github.com/stellar/xdrgen/pull/218
- https://github.com/stellar/rs-stellar-xdr/pull/447

### Merging

Intended to be merged to `main` after:
- #449

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -139,10 +139,15 @@ jobs:
         sys:
         - os: ubuntu-latest
           target: wasm32-unknown-unknown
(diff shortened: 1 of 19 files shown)


dmkozh (approved):
Is this a protocol change? That would be fine, we'll need to release another RC of env, so we could pick this up. But we need to understand that.

PR #458, opened 2025-06-13
Avoid large usage of stack memory

c410-f3r:
Avoids allocating `Arbitrary`-related memory from the stack. `1024 * 10` (`10240`) usually isn't a problem in modern `x86_64` systems so let me know if this change is desired.


c410-f3r:
Let me know if anything is missing

PR #488, opened 2026-01-07
[Experiment] Add sparse XDR decode types

leighmcculloch:
### What

Add code generation for "sparse types" - decode-only XDR types that extract specific nested fields while skipping unneeded data during decoding.

Key changes:
- Add `SkipXdr` trait for skipping over XDR data without allocation
- Add `SeekableSkipXdr` trait for efficient seeking when reader supports `Seek`
- Generate sparse type variants (e.g., `TransactionEnvelopeSparse`, `LedgerCloseMetaTxHashes`)
- Sparse types mirror original structure but only decode specified field paths
- Add benchmarks comparing full vs sparse vs seekable-sparse decoding

### Why

When processing large XDR structures like `LedgerCloseMeta`, applications often only need specific fields (e.g., transaction hashes). Full decoding allocates memory for all fields even when most are discarded.

Sparse decoding provides significant performance improvements:
- **2x faster** with `SkipXdr` (read-based skipping)
- **3.4x faster** with `SeekableSkipXdr` (seek-based skipping)

Benchmarks on real ledger data (1.1MB `LedgerCloseMeta`):
| Method | Time | Throughput |
|--------|------|------------|
| Full decode | 2.30 ms | 473 MiB/s |
| Sparse decode | 1.16 ms | 937 MiB/s |
| Seekable sparse | 0.68 ms | 1.55 GiB/s |

### Example

First instal the `stellar-xdr` cli from this PR:
```
cargo install \
    --locked \
    --git https://github.com/stellar/rs-stellar-xdr \
    --branch sparse-types \
    --features cli \
    stellar-xdr
```

Extract only transaction hashes from a ledger close meta by specifying a type to be code generated that extracts specific fields:

https://github.com/stellar/rs-stellar-xdr/blob/72ba77e446f96949510f1f86f3ad5e172617b1c7/xdr-generator/generate.rb#L38-L46

Decode the value per normal:

```bash
cat ledger.xdr | stellar-xdr decode --type LedgerCloseMeta --input single
```

```
{
  "v2": {
    "ext": {
      "v1": {
        "ext": "v0",
        "soroban_fee_write1_kb": "1000"
      }
    },
    "ledger_header": {
      "hash": "71d4e0bcdc94b8e2e2cb009ef0904c074059e887b7bffbe2ab24fcbcb594cadc",
      "header": {
        "ledger_version": 24,
...
```

Decode the value with the sparse type, only the subset of fields will be decoded:

```bash
cat ledger.xdr | stellar-xdr decode --type LedgerCloseMetaTxHashes --input single
```

```
{
  "v2": {
    "tx_processing": [
      {
        "result": {
          "transaction_hash": "cad07995084476f76f7e5bec6bb7a4a3dc006465063ba2f49e44aa65ea0bdd34"
        }
      },
      {
        "result": {
          "transaction_hash": "8169b61cf900f3017ec09236ee5f9abb1acf2238e8c925e33dd11372f6b367b4"
        }
      },
      {
        "result": {
          "transaction_hash": "1c67ada64a851a2c77fb9b2a455a507172cab3bd148add433a33577e2264465f"
        }
      },
...
```

### Status

This is an experiment and a spike. The implementation explores whether sparse decoding is feasible and beneficial for the XDR library. Further refinement may be needed before production use.

### Prompts

This was built with Claude Opus 4.5, using the following prompts:

```
## Prompt 1: Initial Task (ralph-loop)

Modify the XDR code generation so that in addition to the existing XDR types that get generated, some special types get generated. The special types are specified as the name of an existing type, plus a list of paths to fields, potentially nested through that type and through other types. The code generated for the special type would be decode xdr only, no encoding, and would skip over other fields during the decode process and only keep the fields specified by the paths.

Look at the Makefile to understand how the code generation is kicked off. The generated code is Rust. The code generation script is Ruby. Only modify code in this repository, no code outside of this repository should need modifying.

When complete:
- Code generation should generate the existing XDR types today unaltered.
- Existing tests should pass unaltered.
- Code generation should generate the special types for any configured. Have as a first case a special type for TransactionEnvelope with paths being to the nested 'operations' field.
- New tests exercising a special type decoding should pass.
- README should contain instructions for how to specify more special types.
- Output: <promise>COMPLETE</promise>

## Prompt 2: Feedback Messages (refining the plan)

- "This looks really good. I love that these special types are called 'sparse types'. One thing I'd like to change is that the path should be complete..."

- "I'd like the generated sparse type to preserve the existing structure fully, even nested..."

- "A naming scheme would need to be used for the nested types to avoid collisions..."

- "The test included in step 5 of the plan isn't quite right because it shows the full type as being an enum but the sparse type being a struct..."

- "We should also as a very last step, write a benchmark that tests extracting just the Transaction Hashes..."

- "It is important that the sparse type paths support fields through Vec's. like tx_processing[].result.transaction_hash"

- "why use consume_len in some cases and read_exact in others? Would it be more performant to use consume_len in all cases of skipping?"

## Prompt 3: Read+Seek Variant

lets add to the task list adding a second variant of every sparse type that can accepts a Read+Seek, and use Seek to skip intead of Read. It will be more performant for those types of readers.
```

### DIFF
--- a/benches/sparse_decode.rs
+++ b/benches/sparse_decode.rs
@@ -26,10 +26,10 @@
 use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 
 use stellar_xdr::curr::{
-    AccountId, CreateAccountOp, LedgerCloseMeta, LedgerCloseMetaBatch, LedgerCloseMetaTxHashes,
-    Limits, Memo, MuxedAccount, Operation, OperationBody, Preconditions, PublicKey, ReadXdr,
-    SeekableReadXdr, SequenceNumber, Transaction, TransactionEnvelope, TransactionEnvelopeSparse,
-    TransactionExt, TransactionV1Envelope, Uint256, WriteXdr,
+    AccountId, BufferedReadXdr, CreateAccountOp, LedgerCloseMeta, LedgerCloseMetaBatch,
+    LedgerCloseMetaTxHashes, Limits, Memo, MuxedAccount, Operation, OperationBody, Preconditions,
+    PublicKey, ReadXdr, SeekableReadXdr, SequenceNumber, Transaction, TransactionEnvelope,
+    TransactionEnvelopeSparse, TransactionExt, TransactionV1Envelope, Uint256, WriteXdr,
 };
 
 /// Create a test `TransactionEnvelope` with the specified number of operations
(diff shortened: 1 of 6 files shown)


leighmcculloch:
May revisit after:
- #490

PR #489, opened 2026-01-07
chore: update stellar-strkey

mootz12:
### What

Update the `stellar-strkey` package

### Why

So it can be updated in the soroban-sdk without an old version also being included

### Known limitations

None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ crate-git-revision = "0.0.6"
 
 [dependencies]
 cfg_eval = { version = "0.1.2" }
-stellar-strkey = { version = "0.0.13", optional = true }
+stellar-strkey = { version = "0.0.15", optional = true }
 base64 = { version = "0.22.1", optional = true }
 serde = { version = "1.0.139", features = ["derive"], optional = true }
 serde_with = { version = "3.12.0", features = ["schemars_0_8"], optional = true }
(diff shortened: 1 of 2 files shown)


leighmcculloch (commented):
This looks good, but I think we might keep this for the next major version. As far as I understand, there's nothing critical that's needed. Let me know if I'm missing something.

Cargo.toml
```diff
@@ -20,7 +20,7 @@ crate-git-revision = "0.0.6"
 
 [dependencies]
 cfg_eval = { version = "0.1.2" }
-stellar-strkey = { version = "0.0.13", optional = true }
+stellar-strkey = { version = "0.0.15", optional = true }
```
leighmcculloch:
```suggestion
stellar-strkey = { version = "0.0.16", optional = true }
```

leighmcculloch (commented):
Let's make this change now. This lib is now working towards v27. Let's do the latest strkey version.

leighmcculloch:
We missed the opportunity to do this because the major versions of stellar-xdr and stellar-env-host have already been released, or are we shipping a new stellar-env-host too that'll go into the first v26 stellar-core? cc @dmkozh

leighmcculloch:
Sorry, I got mixed up, this is for v27?

leighmcculloch (approved):

leighmcculloch:
This missed the v27 release, so will need to target v28. And at this point a new version of strkey will be available then, so I'm going to close this.

PR #515, opened 2026-03-16
Add #ifdef cfg support to XDR generator

### LINKED ISSUE #497: Add support for #ifdef #else #endif in .x files
### What

Add support for `#ifdef`/`#ifndef`/`#elif`/`#else`/`#endif` in .x files, mapping them to `#[cfg(feature = "...")` blocks.

### Why

To support the evolution of new features in XDR from the https://github.com/stellar/stellar-xdr repo, without the need for curr/next, something that is coming as a result of:
- https://github.com/stellar/stellar-xdr/issues/295
- https://github.com/stellar/pod-fsr/issues/12

leighmcculloch:
### What:
Lexer tokenizes `#ifdef`, `#else`, and `#endif` preprocessor directives. Parser tracks a cfg condition stack and attaches a `CfgExpr` (`Feature`, `Not`, `All`) to each AST definition inside a preprocessor block, including `#elif`/`#else` branches and nested `#ifdef` blocks; extracted nested definitions inherit the enclosing cfg. Generator propagates cfg from AST definitions to all output structs and introduces `TypeEnumEntry` to carry per-variant cfg alongside the type name. Templates emit `#[cfg(...)]` before type definitions and all generated `impl` blocks when a cfg expression is present. `VARIANTS`/`VARIANTS_STR` and `variants()` in the type enum template change from fixed-size arrays to slices to accommodate a variable number of cfg-gated variants.

### Why:
XDR specification files use C preprocessor directives to conditionally include types based on compile-time features. Without this support, the generator had no way to represent or emit those conditions, making it impossible to produce correct Rust code from specs that rely on `#ifdef` guards.

Close https://github.com/stellar/rs-stellar-xdr/issues/497


sisuresh (approved):

leighmcculloch:
Because this PR has been approved, I'm merging it and split off next changes into:
- #519

PR #518, opened 2026-03-24
Add JSON AST IR with cfg resolution for multi-language XDR code generation

tamirms:
## Summary

Add a `--json-ast` flag to the XDR generator that outputs a language-neutral JSON IR for consumption by code generators in other languages (e.g., Go). Uses shadow types in ir.rs decoupled from the parser's AST, so the IR shape can evolve independently without modifying the parser.

### JSON IR format

- Flat list of definitions, each carrying its computed properties directly
- Internally-tagged encoding: `{"kind": "struct", "name": "Foo", "fixed_size": 12, ...}`
- `fixed_size` only on structs and unions (where it's a non-trivial computation). Enums are always 4 bytes by XDR spec; typedefs are resolved by following the chain — same as consumers do for everything else
- Pre-resolved values: union case idents resolved to integers, named sizes resolved to integers, const values inlined
- Union cases as `[{"value": 0, "name": "TAG_A"}, {"value": 1}]` — single struct with optional name, not parallel arrays
- `resolved_features` metadata when `--feature` is used

### Shadow types (ir.rs)

Dedicated IR types for JSON emission, constructed from the AST via `ir::build()`. This avoids adding serde derives, serde tags, or variant reshaping to the parser's AST. The only AST additions are `CfgExpr::evaluate()` and `UnionArm.name`.

Wire size computation lives in ir.rs (not in the parser) since it's only used by the IR builder. The `ir::build()` function is the single entry point — takes `&XdrSpec` + resolved features, returns `IR`. TypeInfo construction, wire size computation, and const resolution are internal details.

### Cfg resolution (--feature flag)

- `--feature <name>` evaluates `#ifdef`/`#else`/`#endif` and filters the spec
- Comma-delimited for multiple features (consistent with Cargo)
- Removes definitions, union arms, and enum members whose cfg evaluates to false
- Clears cfg annotations on surviving elements
- Restricted to `--json-ast` (Rust codegen uses `#[cfg(...)]` attributes directly)

### Other changes

- `CfgExpr::evaluate()` method on the AST type
- Checked arithmetic in wire size computation
- Snapshot JSON IR to `xdr-definitions-json/xdr.json`
- Union arm `name` field preserved in AST for downstream consumers


xdr-generator-rust/xdr-parser/src/types.rs
```diff
@@ -142,6 +117,158 @@ impl TypeInfo {
...
+// =============================================================================
+// Wire size computation (XDR RFC 4506)
+// =============================================================================
+
+/// Computed properties for each type definition, suitable for serialization into an IR.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct TypeProperties {
+    pub kind: &'static str,
+    /// Total XDR wire size in bytes if the type is fixed-size, None if variable.
+    pub fixed_wire_size: Option<u32>,
```
leighmcculloch: It's not obvious to me how variable sized values size are determined. I guess they are left out, which means your parser has to read the XDR to determine the size. Is that correct?

xdr-generator-rust/xdr-parser/src/types.rs
```diff
@@ -142,6 +117,158 @@ impl TypeInfo {
...
 }
 
+// =============================================================================
+// Wire size computation (XDR RFC 4506)
+// =============================================================================
+
+/// Computed properties for each type definition, suitable for serialization into an IR.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct TypeProperties {
+    pub kind: &'static str,
```
leighmcculloch: Can we use a proper enum for this?

leighmcculloch (commented):
Nice. A few asks inline.

Can we also:
- [ ] Generate the JSON and write it to a new folder at the root, `xdr-definitions-json`? In this repo anything that can be generated gets snapshotted into the repo on every commit. Just do that for the 'curr' branch for the moment.
- [ ] Hold merging this until #515 which is critical path for other work that's happening at the moment.

xdr-generator-rust/xdr-parser/src/types.rs
tamirms: Correct — variable-sized types have fixed_wire_size: null. Their wire size can only be determined by reading the XDR payload at runtime. Only types whose wire size is statically known from the schema get a numeric value.

xdr-generator-rust/xdr-parser/src/ast.rs
```diff
@@ -239,7 +221,8 @@ pub struct Const {
 }
 
 /// XDR type specification.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize)]
+#[serde(tag = "kind")]
```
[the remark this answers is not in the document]
tamirms: I used `#[serde(rename_all = "snake_case")]` on the types. But, I still kept `tag = "kind"` because the JSON IR is designed for consumption by non-Rust code generators (Go today, potentially others). External tags produce two different shapes for the same enum — bare strings for unit variants (`"int"`) and wrapper objects for struct variants (`{"opaque_fixed": {"size": ...}}`). The consumer has to handle both shapes. Internal tags produce uniform objects for every variant (`{"kind": "int"}`, `{"kind": "opaque_fixed", "size": ...}`), which is simpler to unmarshal in Go and is the standard pattern for tagged unions in [TypeScript](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) and other languages.

But, if you still think it's better to remove `tag = "kind"` I can do that and I will just have to write extra unmarshalling logic on the Go side to handle it.

tamirms:
@leighmcculloch can you take another look at the PR now? it is ready for review. I have pulled in the latest changes from master including https://github.com/stellar/rs-stellar-xdr/pull/515 and I have included the xdr-definitions-json fixture file as well

xdr-definitions-json/xdr.json
leighmcculloch, line 1: One thing I noticed about the format is that it requires matching up information from multiple places. There's a separate computed section that you have to lookup fixed size information, when it would be more convenient and intuitive if that was in the definition itself. And the computer const values would also ideally be in the definition of the const too. If they were combined it'd be a list of definitions, and each definition will contain their details, and maybe computed values can be under a computed field of each definition. I think that would be easier to work with. Wdyt?

xdr-generator-rust/generator-definitions-json/src/tests/cfg_filter.rs
leighmcculloch, line 1: If these tests test existing functionality, not new functionality, could we add them in a separate quick pr?

xdr-generator-rust/generator/src/tests/types.rs
```diff
@@ -39,7 +39,7 @@ fn test_base_type_ref_bool() {
 #[test]
 fn test_base_type_ref_opaque_fixed() {
     assert_eq!(
-        base_type_ref(&Type::OpaqueFixed(Size::Literal(32)), None),
+        base_type_ref(&Type::OpaqueFixed { size: Size::Literal { literal: 32 } }, None),
```
leighmcculloch: Regardless of this change using the named fields is a plus I think, it makes this code much more self-descriptive when reading 👍🏻

xdr-definitions-json/xdr.json
tamirms: Yeah, I think that makes a lot of sense. In an earlier conversation you mentioned the idea of using shadow types for the IR. I think that's the right approach here and I've implemented that in the next commit I'm pushing. There will be separate IR types in ir.rs that are purpose-built for JSON emission, decoupled from the parser's AST.

This also means minimal changes to the parser — no serde derives on AST types, no serde tags, no reshaping variants (e.g. tuple to struct) just for serialization. The only AST additions are CfgExpr::evaluate() for feature filtering and a name field on UnionArm.

The shadow types let us build exactly the consumer-friendly format you're describing, plus a few other improvements:

- Each definition carries its computed properties directly (fixed_size, resolved const values) — no separate computed section
- Union case values pre-resolved to integers with optional ident names
- Named sizes (like opaque data[MAX_SIZE]) resolved to integers
- Union arm cases use a single {value, name?} struct instead of parallel arrays

xdr-generator-rust/generator-definitions-json/src/tests/cfg_filter.rs
tamirms: These tests all cover new functionality added in this PR — CfgExpr::evaluate(), filter_spec, ComputedProperties/wire size computation, and the new UnionArm.name field. None of them test pre-existing code.

xdr-generator-rust/xdr-parser/src/ast.rs
```diff
@@ -398,6 +408,9 @@ pub struct UnionDiscriminant {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct UnionArm {
     pub cases: Vec<UnionCase>,
+    /// The declaration name for this arm (e.g., "v0" from "LedgerCloseMetaV0 v0;").
+    /// Empty string for void arms.
+    pub name: String,
```
leighmcculloch: Can we make this an `Option<String>`, and make it `None` for fields without a name like void arms?

leighmcculloch (approved):
I pushed two changes:
* 71376e7 split json ast emitter into separate bin
* 51c7921 use Option for union arm name instead of empty string sentinel which is more idiomatic in Rust

The second one has a small impact on the output, but the first is just shifting the logic into a separate bin.

:shipit:

PR #522, opened 2026-03-30
Gate TypeVariant and Type behind `type_enum` feature

leighmcculloch:
### What
Move `TypeVariant` and `Type` enums out of `src/generated.rs` into a dedicated `src/generated/type_enum.rs` file, gated behind a new `type_enum` cargo feature. Add generator templates (`type_enum.rs.jinja`, `type_enum_definition.rs.jinja`) to produce this file. Enable the `type` feature in `cli`.

### Why
These enums enumerate every XDR type, making them very large. Putting them behind a feature flag lets users who don't need dynamic type dispatch avoid the compile-time and binary-size cost. The majority of our uses of the Rust xdr lib do not require these types.

The places the type enum get used are:
- stellar-cli
- stellar-xdr-cli
- lab

Reduces build time on my own system by about 27%.

### Merging
Targeting v27.


dmkozh (approved):

sisuresh (approved):

leighmcculloch:
Fyi @sisuresh @dmkozh i made changes to the build workflow to get it to pass. The workflow had reached its limited for the number of builds we could do inside a single runner, running out of disk space. We were going to hit this sooner or later, and sooner given the adoption of features. It just happened to trigger on this PR that added a new feature.

This change is largely not consequential to the rest of the pr so I'm going to merge with it without a re-review, but please feel free to take a look and if you have any feedback I'll address it.

PR #528, opened 2026-04-22
generator-definitions-json: make --feature always filter

tamirms:
Fixes the `--feature` flag behavior in `generator-definitions-json` — the binary at `xdr-generator-rust/generator-definitions-json/` that parses `.x` files and emits a JSON AST for downstream code generators. It is invoked by the `make xdr-definitions-json` target to produce the committed snapshot at `xdr-definitions-json/xdr.json`.

### Context

Given an `.x` file like:

```c
struct AccountEntry { ... };     // unconditional

#ifdef CAP_0071
struct NewType { ... };          // gated on cap_0071
#endif
```

### Before this PR

| `generator-definitions-json` invocation | `AccountEntry` | `NewType` |
|-----------------------------------------|----------------|-----------|
| (no flag)                               | kept           | **kept**  |
| `--feature cap_0071`                    | kept           | kept      |

There was no way to get `AccountEntry` *without* `NewType` — no way to express a baseline schema that excludes speculative CAP-gated fields.

### After this PR

| `generator-definitions-json` invocation | `AccountEntry` | `NewType`    |
|-----------------------------------------|----------------|--------------|
| (no flag)                               | kept           | **dropped**  |
| `--feature cap_0071`                    | kept           | kept         |

No flag now means "empty feature set" — `#ifdef X` definitions drop because `X` isn't in the set, unconditional definitions stay. `--feature X,Y` opts into specific CAPs.

### Snapshot

The committed `xdr-definitions-json/xdr.json` regenerates to baseline (no speculative CAPs). This matches `xdr-json/`, which is generated with Cargo's default features and also excludes them.

### DIFF
--- a/xdr-definitions-json/xdr.json
+++ b/xdr-definitions-json/xdr.json
@@ -6807,10 +6807,6 @@
         {
           "name": "ENVELOPE_TYPE_SOROBAN_AUTHORIZATION",
           "value": 9
-        },
-        {
-          "name": "ENVELOPE_TYPE_SOROBAN_AUTHORIZATION_WITH_ADDRESS",
-          "value": 10
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


leighmcculloch (approved):
The change looks fine if the goal is for the tool to always filter, but I thought one of the goals that was expressed about the tool was that it would make it easier for downstream tools to add support for the ifdef feature flags natively, and so the ast json was supposed to contain that information. It seems like it won't do that anymore.

tamirms:
Feature flags are still fully supported — they just live at the tool-invocation boundary rather than inside the IR.

The downstream workflow is: a consumer (e.g., a Go or TypeScript generator) invokes `generator-definitions-json --feature cap_0071` with whatever feature set they care about, gets a JSON IR pre-filtered for that feature set, and generates code from it. New CAPs are reachable exactly this way — no capability is lost by this PR.

Preserving cfg expressions in the IR would only be necessary if we wanted downstream tools to evaluate feature flags *themselves*, at their own codegen time, the way Rust's `#[cfg(feature = "X")]` attributes let Cargo pick at build time. That design is harder to translate to languages without a Cargo-like feature system — e.g., Go build tags don't compose across packages cleanly, so a byte-oriented view type gated behind a tag would force every caller to also be gated. Pre-filtering at IR-generation time sidesteps that problem: the downstream tool never sees cfg metadata, just a concrete schema corresponding to the features it asked for.

So the interface is: the Rust tool always filters to a specific feature set, and downstream consumers pick the feature set they want. That's the workflow this PR codifies.

PR #535, opened 2026-05-27
Add workflow to auto-regenerate XDR from upstream stellar-xdr

sisuresh:
### What

Adds a `Regenerate XDR` workflow (+ `sync-xdr-features.py` helper) that runs daily and on demand (`workflow_dispatch`): it bumps the `xdr` submodule to the latest `stellar/stellar-xdr@main`, regenerates, registers any new feature gates (e.g. `cap_0083`) in `Cargo.toml` and `src/lib.rs`, validates with `make xdr-json` + `cargo test --all-features`, and opens a PR.

### Why

Automates the manual regen so the Rust bindings don't drift behind upstream. Relates to https://github.com/stellar/pod-fsr/issues/5

### Known limitations

The PR is opened with `GITHUB_TOKEN`, so `rust.yml` won't auto-run on it (the workflow self-validates instead; full CI runs on merge). A one-line `token:` swap to a PAT/App token would enable that.


.github/workflows/regenerate-xdr.yml
```diff
@@ -0,0 +1,100 @@
...
+      # Generate the Rust source (no crate compile yet) and record the new xdr-version.
+      - name: Regenerate sources
+        if: steps.xdr.outputs.changed == 'true'
+        run: make clean generate-files xdr-version
+
+      # Register any new XDR feature gates (e.g. a newly added cap_XXXX) in Cargo.toml and
+      # src/lib.rs, so the crate compiles and reports the gate. No-op when none are new.
+      - name: Register new XDR feature flags
+        if: steps.xdr.outputs.changed == 'true'
+        run: python3 .github/scripts/sync-xdr-features.py
```
dmkozh, line 55: Don't we also need to remove the old feature flags?

sisuresh: Good point. Fixed.

dmkozh (approved):

PR #547, opened 2026-06-17
Update stellar-strkey to 0.0.18

leighmcculloch:
### What

Bump the `stellar-strkey` dependency from 0.0.13 to 0.0.18 and adapt the hand-written strkey usage in `src/str.rs` to the 0.0.17+ API.

### Why

0.0.18 is the stellar-strkey release line for protocol 28. The upgrade requires a few source changes because the API differs from 0.0.13: the `Strkey::PrivateKeyEd25519` variant was removed, and `ed25519::SignedPayload` now has private fields exposed via `new(ed25519, payload)`, `ed25519()`, and `payload()` accessors instead of public struct fields. `src/str.rs` was updated to construct and read `SignedPayload` through these accessors and to drop the removed `PrivateKeyEd25519` match arms. The lockfile was updated to pin 0.0.18.

This supersedes the earlier 0.0.17 bump: 0.0.17 contained a CLI bug (`Decoded<Strkey>` rejected owned JSON keys, breaking `strkey encode`), fixed in 0.0.18 by stellar/rs-stellar-strkey#125.

### Known limitations

N/A

---

## Protocol 28 coordination

⚠️ This change targets versions of software that will ship for **Protocol 28**, and should **not** be merged unless this repository is ready to accept Protocol 28 changes.

This is one of a coordinated set of PRs bumping `stellar-strkey` to **0.0.18** across the Protocol 28 component stack. It is preferred that all Protocol 28 releases of these components depend on the **same** version of `stellar-strkey`:

- stellar/rs-stellar-xdr#547
- stellar/rs-soroban-env#1694
- stellar/rs-soroban-sdk#1913
- stellar/rs-stellar-rpc-client#99
- stellar/stellar-cli#2614

`stellar-strkey` 0.0.18 raises the minimum supported Rust version to 1.87 and updates/adds transitive dependencies (`heapless` 0.9, `zeroize`), so MSRV and supply-chain policy (cargo-deny / cackle) updates may be required in some repositories before CI is green.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Stellar Development Foundation <info@stellar.org>"]
 license = "Apache-2.0"
 version = "27.0.0"
 edition = "2021"
-rust-version = "1.84.0"
+rust-version = "1.87.0"
 
 [[bin]]
 name = "stellar-xdr"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ crate-git-revision = "0.0.6"
 
 [dependencies]
 cfg_eval = { version = "0.1.2", optional = true }
-stellar-strkey = { version = "0.0.17", optional = true }
+stellar-strkey = { version = "0.0.18", optional = true }
 base64 = { version = "0.22.1", optional = true }
 serde = { version = "1.0.139", features = ["derive"], optional = true }
 serde_with = { version = "3.12.0", features = ["schemars_0_8"], optional = true }
@@ -66,3 +66,9 @@ cli = ["std", "type_enum", "base64", "serde", "serde_json", "schemars", "arbitra
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docs"]
+
+# Temporary: stellar-strkey 0.0.18 is not yet published to crates.io. Point at
+# the release/v0.0.18 branch so this PR can be validated against the actual
+# 0.0.18 code ahead of the release. Remove once 0.0.18 is published.
+[patch.crates-io]
+stellar-strkey = { git = "https://github.com/stellar/rs-stellar-strkey", rev = "73bf43778a4acad5e7b3589f54a5ec3607c4b7b1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ crate-git-revision = "0.0.6"
 
 [dependencies]
 cfg_eval = { version = "0.1.2", optional = true }
-stellar-strkey = { version = "0.0.13", optional = true }
+stellar-strkey = { version = "0.0.17", optional = true }
 base64 = { version = "0.22.1", optional = true }
 serde = { version = "1.0.139", features = ["derive"], optional = true }
 serde_with = { version = "3.12.0", features = ["schemars_0_8"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,9 +66,3 @@ cli = ["std", "type_enum", "base64", "serde", "serde_json", "schemars", "arbitra
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docs"]
-
-# Temporary: stellar-strkey 0.0.18 is not yet published to crates.io. Point at
-# the release/v0.0.18 branch so this PR can be validated against the actual
-# 0.0.18 code ahead of the release. Remove once 0.0.18 is published.
-[patch.crates-io]
-stellar-strkey = { git = "https://github.com/stellar/rs-stellar-strkey", rev = "73bf43778a4acad5e7b3589f54a5ec3607c4b7b1" }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,8 +6,6 @@
 #![allow(clippy::tabs_in_doc_comments)]
 #![allow(clippy::doc_markdown)]
 #![allow(clippy::doc_lazy_continuation)]
-// Pedantic lints raised by the newer clippy that ships with the bumped MSRV,
-// firing on the auto-generated code.
 #![allow(clippy::unnecessary_semicolon)]
 #![allow(clippy::non_std_lazy_statics)]
 #![allow(unused_attributes)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,10 @@
 #![allow(clippy::tabs_in_doc_comments)]
 #![allow(clippy::doc_markdown)]
 #![allow(clippy::doc_lazy_continuation)]
+// Pedantic lints raised by the newer clippy that ships with the bumped MSRV,
+// firing on the auto-generated code.
+#![allow(clippy::unnecessary_semicolon)]
+#![allow(clippy::non_std_lazy_statics)]
 #![allow(unused_attributes)]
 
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -6,6 +6,10 @@
 #![allow(clippy::tabs_in_doc_comments)]
 #![allow(clippy::doc_markdown)]
 #![allow(clippy::doc_lazy_continuation)]
+// Pedantic lints raised by the newer clippy that ships with the bumped MSRV,
+// firing on the auto-generated code.
```
leighmcculloch:
```suggestion
```

src/str.rs
```diff
@@ -204,10 +203,8 @@ impl core::fmt::Display for SignerKeyEd25519SignedPayload {
             ed25519: Uint256(ed25519),
             payload,
         } = self;
-        let k = stellar_strkey::ed25519::SignedPayload {
-            ed25519: *ed25519,
-            payload: payload.into(),
-        };
+        let k = stellar_strkey::ed25519::SignedPayload::new(*ed25519, payload.as_ref())
+            .map_err(|_| core::fmt::Error)?;
```
leighmcculloch, line 207: There's a subtle meaningful change here that I'd like to give a heads up and a chance for commentary before this is merged.

The `Display::fmt` implementation is called by `ToString::to_string`. `Display::fmt` can return an error, but `ToString::to_string` doesn't and so panics if `fmt` returns an error.

Previously you could construct a `SignerKeyEd25519SignedPayload` with a zero length payload all the way up to a `u32::MAX` length payload and it would render a string. It would panic in payloads longer than `u32::MAX`. However, that string generated might be invalid, since a payload must be 1 to 64 bytes long in stellar-core, we implemented the rs-stellar-strkey parse impl to always require within those lengths. Attempts to then parse the key would error.

The latest rs-stellar-strkey rectified the asymmetry in its own type for two reasons:
1. The asymmetry was identified as a potential security issue that an application could be impacted by if they used the type naively.
2. To support no-std no-alloc types a low max length was convenient.

This has a side-effect that it's possible to construct a `SignerKeyEd25519SignedPayload` xdr type whose `Display` impl returns an error and its `ToString` impl panics. It was possible before if the payload was `u32::MAX+1+`, but now it is possible with `0` and `65+`

I think this change is inevitable unless we remove the limits from rs-stellar-strkey, and walked back the no-std no-alloc support for that type.

I don't think this can effect stellar-core, but there is some XDR, albeit invalid, that cannot be represented in string and json forms.

cc @graydon @dmkozh @mootz12

dmkozh: I don't understand why do we need to panic in the Display impl. Can't we just return something like `<invalid>`? While this may not necessarily affect env today, I can imagine code that logs the invalid payload if it fails some kind of validation - it would be weird for the logging logic to cause panic then.

leighmcculloch: It will create the same problem that the auditors had with the strkey library, that the type will not roundtrip which can also be a failure point similar to the failure point we are trying to address here. Maybe it's more appropriate to use the `Debug` impl in that case.

leighmcculloch: I could look at introducing a validation into the XDR type so that it cannot be constructed with a payload outside the range. We do that today with max lengths on XDR var arrays, but not min lengths.

dmkozh: > , that the type will not roundtrip

Why is that an issue, if the input is invalid? IMO it doesn't make sense to require every input to be able to roundtrip. Moreover, this is about XDR types, not strkey types. There is no requirement for XDR types to roundtrip through Display, is there? It's not a serialization format.

> I could look at introducing a validation into the XDR type so that it cannot be constructed with a payload outside the range.

That's an option, but I'm still not sold on the roundtrip argument TBH.

leighmcculloch: > There is no requirement for XDR types to roundtrip through Display, is there? It's not a serialization format.

Historically the `Display` impl has roundtripped with the `FromStr` impl for XDR types that implement it, and the XDR-JSON impl uses both, so they do need to round trip.

leighmcculloch: > \<invalid\>

I think something like this, that is obviously invalid, is probably the best we can do. Maybe just capitalise it for some similarity to the normal output:

`<INVALID>`

leighmcculloch: Actually, it may be helpful to embed the `Debug` impl too, so:

`<INVALID:{Debug}>`

That way the output captures the data and can be inspected.

Cargo.toml
```diff
@@ -20,7 +20,7 @@ crate-git-revision = "0.0.9"
 
 [dependencies]
 cfg_eval = { version = "0.1.2", optional = true }
-stellar-strkey = { version = "0.0.13", optional = true }
+stellar-strkey = { version = "0.0.18", optional = true }
```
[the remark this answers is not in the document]
sisuresh: fixed

src/str.rs
```diff
@@ -204,9 +203,12 @@ impl core::fmt::Display for SignerKeyEd25519SignedPayload {
...
         } = self;
-        let k = stellar_strkey::ed25519::SignedPayload {
-            ed25519: *ed25519,
-            payload: payload.into(),
+        // A strkey signed payload must be 1 to 64 bytes. The XDR type caps the
+        // payload at 64 bytes, so the only payload that cannot be rendered as
+        // a strkey is an empty one. Render it as obviously invalid instead of
+        // returning an error, which would cause to_string to panic.
+        let Ok(k) = stellar_strkey::ed25519::SignedPayload::new(*ed25519, payload.as_ref()) else {
+            return f.write_str("<INVALID:SignerKeyEd25519SignedPayload:EMPTY_PAYLOAD>");
```
leighmcculloch: Re the Codex comment on empty payloads: `Display` no longer returns an error for them. The XDR type caps the payload at 64 bytes, so the only payload stellar-strkey rejects is an empty one, and it now renders as `<INVALID:SignerKeyEd25519SignedPayload:EMPTY_PAYLOAD>`, so `to_string` doesn't panic. Covered by tests in `tests/str.rs` (ec506f5).

---

src/str.rs
leighmcculloch: > > I could look at introducing a validation into the XDR type so that it cannot be constructed with a payload outside the range.
> 
> That's an option, but I'm still not sold on the roundtrip argument TBH.

For consideration:
- https://github.com/stellar/rs-stellar-xdr/pull/591

leighmcculloch: Thinking about the two options:

The benefit of `<INVALID>` is that the libraries mostly behave the same way they do today. Today you can turn XDR into XDR-JSON that can't roundtrip back if the signed payload zero length. Instead of a P address that's actually invalid, it will be the INVALID string. So you can't generate XDR from invalid XDR-JSON. XDR can generate invalid XDR-JSON.

The benefit of the min length for BytesM is that nothing can generate XDR from invalid bytes or invalid XDR-JSON. And XDR cannot generate invalid XDR-JSON.

If this was a more important XDR type then I think the min length would be worth doing, given this is a type that if we could we'd probably remove from the protocol, we can just run with 1.

leighmcculloch: I've updated this PR so the output is now:
`<INVALID:GA7QYNF7SOWQ3GLR2BGMZEHXAVIRZA4KVWLTJJFC7MGXUA74P7UJVSGZ:EMPTY_PAYLOAD>`

sisuresh (approved):

PR #557, opened 2026-07-20
error on unknown fields in encode command JSON input

leighmcculloch:
### What

Add a `serde_ignored` feature that enables `from_json_collecting_ignored_fields` and `deserialize_json_collecting_ignored_fields` on the `Type` enum. The `cli` feature enables `serde_ignored`, and the `encode` command uses these methods to detect unknown JSON fields and exit non-zero with an error message to stderr when any are found.

### Why

The `encode` command silently ignores extra fields in the JSON input (e.g. typos or fields from a different type) and produces XDR as if they weren't there. This masks mistakes. Closes #554.

### Known limitations

This has some limitations wrt the custom deserialiers that get used on some types, so this will improve the use of the encode cli command for most types but not a couple of the types like the integer types that parse both Number and String inputs. This is still a net benefit regardless.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,8 +62,10 @@ arbitrary = ["std", "dep:arbitrary"]
 hex = []
 rand = ["dep:rand"]
 
+serde_ignored = ["std", "serde_json", "dep:serde_ignored"]
+
 # Features for the CLI.
-cli = ["std", "type_enum", "base64", "serde", "serde_json", "schemars", "arbitrary", "rand", "dep:clap", "dep:thiserror", "dep:ethnum", "dep:serde_ignored"]
+cli = ["std", "type_enum", "base64", "serde", "serde_json", "schemars", "arbitrary", "rand", "dep:clap", "dep:thiserror", "dep:ethnum", "serde_ignored"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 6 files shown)


src/cli/encode.rs
```diff
@@ -101,14 +110,24 @@ macro_rules! run_x {
                 match self.input_format {
                     InputFormat::Json => match self.output_format {
                         OutputFormat::Single => {
-                            let t = crate::Type::from_json(r#type, f)?;
+                            let mut de =
+                                serde_json::Deserializer::new(serde_json::de::IoRead::new(f));
+                            let (t, ignored) = crate::Type::deserialize_json_with_ignored_fields(
+                                r#type, &mut de,
+                            )?;
+                            check_ignored_fields(ignored)?;
```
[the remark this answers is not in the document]
leighmcculloch: Good catch -- added a call to `de.end()` after deserialization in both the `Single` and `SingleBase64` paths to restore the trailing-content validation that `from_reader` provided.

Cargo.toml
```diff
@@ -60,9 +61,10 @@ schemars = ["alloc", "serde", "serde_json", "dep:schemars"]
 arbitrary = ["std", "dep:arbitrary"]
 hex = []
 rand = ["dep:rand"]
+serde_ignored = ["std", "serde_json", "dep:serde_ignored"]
```
sisuresh, line 64: From a claude review - 

> serde_ignored feature is a no-op unless type_enum is also on. The new methods live inside the #[cfg(feature = "type_enum")] impl Type block, gated additionally on serde_ignored. Enabling only serde_ignored compiles the dependency but adds no reachable API. Consider serde_ignored = ["type_enum", "serde_json", "dep:serde_ignored"] so the feature is self-consistent. (cli enables both, so the CLI is unaffected — this only matters for a downstream enabling the feature directly.)

leighmcculloch: I agree but also don't love turning on type enum just because serde_ignored is on, it's not exactly intuitive, I don't think that actually solves a problem, it creates a different type of problem.

sisuresh (approved):

PR #569, opened 2026-08-15
fix: reject AssetCode12 JSON shorter than 5 characters

### LINKED ISSUE #432: xdrjson: AssetCode12 allows invalid JSON value that doesn't round trip
### What version are you using?

v22.1.0

### What did you do?

```
$ echo '"ABC"' \
    | stellar xdr encode --type AssetCode12 \
    | stellar xdr decode --type AssetCode12
"ABC\\0\\0"
```

### What did you expect to see?

Expected that the call to encode would fail, and that `"ABC"` would be considered an invalid `AssetCode12`, since that type always maps to at least 5 characters in JSON so as to clearly distinguish with `AssetCode4`.

The expectation is that only `"ABC\\0\\0"` would round trip for `AssetCode12` for this value.

XDR-JSON is intended to be like XDR, that there is only a single representation for each XDR value.

### What did you see instead?

The inconsistent round-tripping.

Galmanus:
### What
`AssetCode12::from_str` now rejects inputs that unescape to fewer than 5 bytes, returning `Error::Invalid`.

### Why
Closes #432. `AssetCode12::from_str` discarded the byte count and accepted any value, including sub-5-byte codes like `"abc"`. That value does not round-trip: the `Display` impl always renders at least 5 characters (any shorter code is an `AssetCode4`), so `"abc"` parsed to `[a,b,c,0,…]` and re-serialized as `"abc\0\0"`.

This is the read-side counterpart to #358 ("Fix render of AssetCode12 to JSON when shorter than 5 chars"), which added the 5-character minimum to `Display` but left `FromStr` unchanged. With both sides aligned, XDR<>JSON round-trips for `AssetCode12`.

### Scope
- One hand-written function in `src/str.rs` (`impl FromStr for AssetCode12`); no codegen / `generated.rs` regeneration.
- The `AssetCode` enum's `FromStr` has independent length routing (`n <= 4` → `AssetCode4`, else `AssetCode12`) and does not call `AssetCode12::from_str`, so it is unaffected.
- The serde `DeserializeFromStr` derive on `AssetCode12` goes through `FromStr`, so the CLI `encode` path (the issue's repro) is fixed transitively.

### Tests
`tests/str.rs::asset_code_12_from_str` updated: the five sub-5-byte cases (`""`/`"a"`/`"ab"`/`"abc"`/`"abcd"`) now assert `Err(Error::Invalid)` (they asserted the buggy `Ok(...)` before), plus a new round-trip assertion for the canonical 5-byte short form. `cargo test --test str` (50 tests) and `cargo clippy` are green.

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -405,8 +405,8 @@ impl core::str::FromStr for AssetCode12 {
         // the XDR<>JSON conversion round-trippable.
         const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
-            .map_err(|_| Error::Invalid)?;
+        let n =
+            escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
         if n < MIN_LENGTH {
             return Err(Error::Invalid);
         }
--- a/src/str.rs
+++ b/src/str.rs
@@ -394,6 +394,10 @@ impl core::fmt::Display for AssetCode4 {
 
 impl core::str::FromStr for AssetCode12 {
     type Err = Error;
+    /// Parses an `AssetCode12`. Requires at least 5 characters: any code shorter
+    /// than 5 is an `AssetCode4`, and accepting it here would not round-trip (the
+    /// `Display` impl always renders at least 5 characters). For length-agnostic
+    /// parsing that selects the right variant, use `AssetCode`'s `FromStr`.
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
         // AssetCode12's are always at least 5 characters, because any asset
         // code shorter than 5 characters is an AssetCode4. This mirrors the
--- a/src/str.rs
+++ b/src/str.rs
@@ -394,7 +394,7 @@ impl core::fmt::Display for AssetCode4 {
 
 impl core::str::FromStr for AssetCode12 {
     type Err = Error;
-    /// Parses an `AssetCode12`. Requires at least 5 characters: any code shorter
+    /// Parses an `AssetCode12`. Requires at least 5 bytes: any code shorter
     /// than 5 is an `AssetCode4`, and accepting it here would not round-trip (the
     /// `Display` impl always renders at least 5 characters). For length-agnostic
     /// parsing that selects the right variant, use `AssetCode`'s `FromStr`.
--- a/src/str.rs
+++ b/src/str.rs
@@ -399,7 +399,7 @@ impl core::str::FromStr for AssetCode12 {
     /// `Display` impl always renders at least 5 characters). For length-agnostic
     /// parsing that selects the right variant, use `AssetCode`'s `FromStr`.
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
-        // AssetCode12's are always at least 5 characters, because any asset
+        // An AssetCode12 is always at least 5 characters, because any asset
         // code shorter than 5 characters is an AssetCode4. This mirrors the
         // Display impl, which always renders at least 5 characters, and keeps
         // the XDR<>JSON conversion round-trippable.
--- a/src/str.rs
+++ b/src/str.rs
@@ -395,8 +395,17 @@ impl core::fmt::Display for AssetCode4 {
 impl core::str::FromStr for AssetCode12 {
     type Err = Error;
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
+        // AssetCode12's are always at least 5 characters, because any asset
+        // code shorter than 5 characters is an AssetCode4. This mirrors the
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
+            .map_err(|_| Error::Invalid)?;
+        if n < MIN_LENGTH {
+            return Err(Error::Invalid);
+        }
         Ok(code)
     }
 }
--- a/tests/str.rs
+++ b/tests/str.rs
@@ -589,10 +589,7 @@ fn asset_code_12_from_str() {
     assert_eq!(AssetCode12::from_str("abcd"), Err(Error::Invalid));
     // The canonical 5-byte escaped short form is accepted and round-trips with
     // Display (see asset_code_12_to_string).
-    assert_eq!(
-        AssetCode12::from_str(r"abc\0\0"),
-        Ok(AssetCode12(*b"abc\0\0\0\0\0\0\0\0\0"))
-    );
+    assert_eq!(AssetCode12::from_str(r"abc\0\0"), Ok(AssetCode12(*b"abc\0\0\0\0\0\0\0\0\0")));
     assert_eq!(AssetCode12::from_str("abcde"), Ok(AssetCode12(*b"abcde\0\0\0\0\0\0\0")));
     assert_eq!(AssetCode12::from_str("abcdef"), Ok(AssetCode12(*b"abcdef\0\0\0\0\0\0")));
     assert_eq!(AssetCode12::from_str("abcdefg"), Ok(AssetCode12(*b"abcdefg\0\0\0\0\0")));
--- a/tests/str.rs
+++ b/tests/str.rs
@@ -580,11 +580,19 @@ fn asset_code_4_to_string() {
 #[test]
 #[rustfmt::skip]
 fn asset_code_12_from_str() {
-    assert_eq!(AssetCode12::from_str(""), Ok(AssetCode12(*b"\0\0\0\0\0\0\0\0\0\0\0\0")));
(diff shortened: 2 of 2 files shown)


src/str.rs
```diff
@@ -395,8 +395,17 @@ impl core::fmt::Display for AssetCode4 {
...
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
+            .map_err(|_| Error::Invalid)?;
+        if n < MIN_LENGTH {
+            return Err(Error::Invalid);
+        }
```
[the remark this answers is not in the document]
Galmanus: Thanks — the breaking-change flag is fair, so here is the reasoning. The change removes behavior that did not round-trip: `AssetCode12::from_str` accepted sub-5-byte inputs, and since the `Display` impl (fixed in #358) always renders at least 5 characters, `"abc"` parsed to `[a,b,c,0,0]` and serialized back as `"abc\\0\\0"`. Anything shorter than 5 characters is an `AssetCode4`, so accepting it as `AssetCode12` was ambiguous by construction.

On the explicit non-ambiguous parser: that already exists. `impl FromStr for AssetCode` routes by length (`n <= 4` -> `AssetCode4`, else `AssetCode12`) and does not call `AssetCode12::from_str`, so it is unaffected — callers who want length-agnostic parsing should use `AssetCode`. This PR only makes the direct `AssetCode12` constructor consistent with that enum's invariant and with the `Display` fix from #358.

Good call on documenting it — I added a doc comment on `AssetCode12::from_str` stating the >=5-character invariant and pointing to `AssetCode` for length-agnostic parsing.

src/str.rs
```diff
@@ -395,8 +395,17 @@ impl core::fmt::Display for AssetCode4 {
...
+        // AssetCode12's are always at least 5 characters, because any asset
+        // code shorter than 5 characters is an AssetCode4. This mirrors the
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
+            .map_err(|_| Error::Invalid)?;
+        if n < MIN_LENGTH {
```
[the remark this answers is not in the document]
Galmanus: Kept `n` for consistency with the sibling `impl FromStr for AssetCode` in this same file, which uses `n` for the decoded byte count returned by `unescape_into`. Happy to rename both to `len`/`byte_len` in one pass if you'd prefer the more descriptive name across the module.

src/str.rs
```diff
@@ -395,8 +395,17 @@ impl core::fmt::Display for AssetCode4 {
 impl core::str::FromStr for AssetCode12 {
     type Err = Error;
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
+        // AssetCode12's are always at least 5 characters, because any asset
```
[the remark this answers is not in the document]
Galmanus: Fixed in 6649b06 — dropped the apostrophe (now "An AssetCode12 is always at least 5 characters").

src/str.rs
```diff
@@ -394,9 +394,22 @@ impl core::fmt::Display for AssetCode4 {
...
+    /// parsing that selects the right variant, use `AssetCode`'s `FromStr`.
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
+        // An AssetCode12 is always at least 5 characters, because any asset
+        // code shorter than 5 characters is an AssetCode4. This mirrors the
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
```
leighmcculloch: `cargo fmt --all --check` fails on this hand-wrapped statement, so the `fmt` CI job (and the `complete` gate) will be red. rustfmt wants the break after `let n =`, not before `.map_err`:

```rust
        let n =
            escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
```

Verified by running `rustfmt --edition 2021 --check` against this branch's `src/str.rs`.

---

src/str.rs
```diff
@@ -394,9 +394,22 @@ impl core::fmt::Display for AssetCode4 {
 
 impl core::str::FromStr for AssetCode12 {
     type Err = Error;
+    /// Parses an `AssetCode12`. Requires at least 5 characters: any code shorter
```
leighmcculloch: The doc comment (and the inline comment below) says "at least 5 characters", but the guard counts *unescaped bytes* — `unescape_into` returns a byte count. The two differ for non-ASCII and for escapes:

- `AssetCode12::from_str("héll")` — 4 characters, 5 bytes → `Ok`
- `AssetCode12::from_str("abcd")` — 4 characters, 4 bytes → `Err`

The PR description's phrasing ("unescape to fewer than 5 bytes") is accurate; worth using it here too, e.g. "Requires at least 5 bytes after unescaping". Same for the `Display` claim on the next line — it always renders at least 5 *bytes*, which may be more than 5 characters once escaped.

---

src/str.rs
```diff
@@ -394,9 +394,22 @@ impl core::fmt::Display for AssetCode4 {
...
+        // An AssetCode12 is always at least 5 characters, because any asset
+        // code shorter than 5 characters is an AssetCode4. This mirrors the
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
+            .map_err(|_| Error::Invalid)?;
+        if n < MIN_LENGTH {
```
leighmcculloch: Worth flagging as a semver-breaking change rather than a bug fix: this narrows what public `FromStr` — and therefore `Deserialize` — accepts. `Cargo.toml` currently sits on `28.0.0`, which is already tagged, so if this lands in a `28.0.1`/`28.1.0` patch, downstream code that today parses JSON like `{"credit_alphanum12":{"asset_code":"USD"}}` would silently start getting `Err(Error::Invalid)` on a non-major bump.

The behaviour is correct per protocol (a <5-byte code is an `AssetCode4`), so no objection to the change itself — just that it wants a major release and a release note. Round-trip is safe: `Display` for `AssetCode12` always emits ≥5 bytes escaped, so every XDR-derived string still unescapes to ≥5 bytes and XDR→JSON→XDR stays lossless, including all-zero and sparse codes like `[0,…,0,x]`. `AssetCode`'s union `FromStr` routes `n <= 4` to `AssetCode4` and never reaches here, so it's unaffected.

---

src/str.rs
leighmcculloch:
```suggestion
    /// Parses an `AssetCode12`. Requires at least 5 bytes: any code shorter
```

tests/str.rs
```diff
@@ -580,11 +580,19 @@ fn asset_code_4_to_string() {
...
+    assert_eq!(AssetCode12::from_str("a"), Err(Error::Invalid));
+    assert_eq!(AssetCode12::from_str("ab"), Err(Error::Invalid));
+    assert_eq!(AssetCode12::from_str("abc"), Err(Error::Invalid));
+    assert_eq!(AssetCode12::from_str("abcd"), Err(Error::Invalid));
+    // The canonical 5-byte escaped short form is accepted and round-trips with
+    // Display (see asset_code_12_to_string).
+    assert_eq!(
+        AssetCode12::from_str(r"abc\0\0"),
+        Ok(AssetCode12(*b"abc\0\0\0\0\0\0\0\0\0"))
+    );
```
leighmcculloch:
```suggestion
    assert_eq!(AssetCode12::from_str(r"abc\0\0"), Ok(AssetCode12(*b"abc\0\0\0\0\0\0\0\0\0")));
```

leighmcculloch (approved):
Thanks for submitting this fix. Couple minor tweaks.

src/str.rs
```diff
@@ -394,9 +394,22 @@ impl core::fmt::Display for AssetCode4 {
...
     fn from_str(s: &str) -> core::result::Result<Self, Self::Err> {
+        // An AssetCode12 is always at least 5 characters, because any asset
+        // code shorter than 5 characters is an AssetCode4. This mirrors the
+        // Display impl, which always renders at least 5 characters, and keeps
+        // the XDR<>JSON conversion round-trippable.
+        const MIN_LENGTH: usize = 5;
         let mut code = AssetCode12([0u8; 12]);
-        escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
+        let n = escape_bytes::unescape_into(&mut code.0, s.as_bytes())
+            .map_err(|_| Error::Invalid)?;
```
leighmcculloch:
```suggestion
        let n =
            escape_bytes::unescape_into(&mut code.0, s.as_bytes()).map_err(|_| Error::Invalid)?;
```

PR #580, opened 2026-09-08
Add const XDR serialization

leighmcculloch:
> [!TIP]
> **Reviewing Tips**
>
> I would focus a review of this change on two areas:
> - Sampling a few of the `const::{Type}` generated types to get an understanding for what code is being generated.
> - The new `ConstWriter` type and the primitive XDR encoding methods.

> [!NOTE]
> Part of a stack of PRs that must merge in this order.
>
> A first group of PRs change contract specs so that they are produced at compile time instead of at proc-macro execution time. This provides the foundation to construct the specs from information that is only known at compile time, like the fully qualified name of a type:
> 1. stellar/rs-stellar-xdr#580 ← this PR
> 1. stellar/rs-soroban-sdk#2061
> 1. stellar/rs-soroban-sdk#1965
>
> A second group of PRs changes the type names that the sdk stores in specs are fully qualified type names, and then the cli reduces them down to unique simple identifiers. During the contract build the types are given names like `::mycrate::mymod::MyType` instead of `MyType`. Then the cli reduces them back down to simple names after spec shaking. Qualified type names make it possible to uniquely identify types in the spec, even when they have the same name. This resolves several problems with contract specs the type identify problem (https://github.com/stellar/rs-soroban-sdk/issues/1570), type aliases limitations (https://github.com/stellar/rs-soroban-sdk/issues/1857 https://github.com/stellar/rs-soroban-sdk/issues/1063), and optimise spec shaking data section size (https://github.com/stellar/rs-soroban-sdk/issues/1978):
> 1. stellar/stellar-xdr#312
> 1. stellar/rs-stellar-xdr#543
> 1. stellar/rs-soroban-sdk#1970
> 1. stellar/stellar-cli#2674
>
> Note that downstream clients and SDKs should see no, or little, change because the cli will during the build process reduce the fully qualified names back to simple unique names. Contracts that had colliding type names, which meant they could not be used with clients, will now work.
>
> A third group of PRs are an optimisation to spec shaking v2, and will use the new unique type names to shake type specs by reachability, producing a dependency graph (thanks @mootz12), so that only spec entries that can't be reached from fns, like errors and events, get spec markers from dead-code-elimination:
> 1. stellar/rs-soroban-sdk#2043
> 1. stellar/stellar-cli#2720

### What

Add a `const` feature giving every generated type an alternative XDR encoding that can be used in const contexts. Add a static-lifetime borrowing `const::{Type}` counterpart for every generated type that directly or transitively contains heap data.

### Why

The owned types require heap allocation to construct, so XDR values can't be built from borrowed slices inside const contexts. The const variant types can be constructed in const contexts.

XDR encoding is used by soroban-sdk to encode contract specs during proc-macro execution. The soroban-sdk will be changing when it encodes contracts specs to during compilation so that the specs can contain information only available during compile.

### Known limitations

The `const::{Type}` types are locked down only for use with borrowed data from static lifetimes, which isn't entirely necessary. An earlier implementation I did in stellar/rs-stellar-xdr#560 and stellar/rs-stellar-xdr#562 had these borrowed types as `{Type}Ref<'a>` where the lifetime was configurable and the type looked like it was useful as a ref type. However, in practice, because of the nature that these ref types get nested within arrays inside other ref types, they are not convertible to and from owned types and not really useful outside of a const static lifetime context. For that reason I've kept the types hyper focused on the immediate use case. If we ever have a use case for a more general ref variant of generated types we can reexplore stellar/rs-stellar-xdr#560.

The const implementation is a separate implementation because a const implementation cannot use traits, which means it can't stream to a writer. It is not practical to maintain a single implementation that offers the same features of both. So instead there are two implementations and a test ensures they work identically.

I would have really liked to fuzz test the const xdr encoding with the existing xdr encoding, however I don't see a way to do that. We can't convert from owned to const types because of the way the const types are structured. Even if they were ref types with lifetimes we can't generate ref types from arbitrary because there would be nothing to own the values. Open to feedback on this, it's possible a slightly different way of architecting the ref types might make this possible.

### Other Changes

1. I slipped a fix in here to the existing VecM, BytesM, StringM type Arbitrary implementations. All three could result in the types exceeding their max lengths. When I wrote the const variants of the types, they didn't have the same bug, and so I needed to fix the other for consistency.

2. I modified the `stellar-xdr generate arbitrary` cli command so that it can accept entropy from a cargo-fuzz corpus file, which is a really helpful way to inspect and view what XDR values the corpus accepts.


xdr-generator-rust/generator/templates/const_to_xdr.rs.jinja
```diff
@@ -0,0 +1,38 @@
...
+impl {{ recv }} {
+    /// The exact XDR-encoded length of this value, in bytes.
+    ///
+    /// Evaluable in a const context, so a caller (such as a proc-macro) can
+    /// size a buffer for [`Self::const_to_xdr`] at compile time.
+    #[must_use]
+    pub const fn const_xdr_len(&self) -> usize {
+        let mut empty: [u8; 0] = [];
+        let mut w = ConstWriter::new(&mut empty);
+        w.{{ write_fn }}(self);
```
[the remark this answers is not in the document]
leighmcculloch: The `generate_to_file` fn was removed in:
- #578 

The `generate` fn is being removed in:
- #582

The reason they have been or are being removed is that the xdr lib doesn't use either anymore for generating itself, as it uses `generate_to_dir` and writes out multiple files instead.

While Copilot is correct that there's a gap here, I don't want to invest fixing it in this PR when that code is dead code and is being removed separately.

src/generated/account_entry.rs
```diff
@@ -100,3 +100,73 @@ impl WriteXdr for AccountEntry {
         })
     }
 }
+
+/// AccountEntryConst is a borrowing equivalent of [`AccountEntry`] over `'static`
+/// data, for const XDR encoding.
+#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AccountEntryConst {
```
dmkozh: Why is the struct itself not guarded with `const` feature?

dmkozh: I feel like suffixing types with `const` might be somewhat confusing, maybe prefix would make more sense, so that this is clearly separated from non-const types?

src/generated/account_entry.rs
```diff
@@ -100,3 +100,73 @@ impl WriteXdr for AccountEntry {
...
+}
+
+#[cfg(feature = "const")]
+impl AccountEntryConst {
+    /// The exact XDR-encoded length of this value, in bytes.
+    ///
+    /// Evaluable in a const context, so a caller (such as a proc-macro) can
+    /// size a buffer for [`Self::const_to_xdr`] at compile time.
+    #[must_use]
+    pub const fn const_xdr_len(&self) -> usize {
```
dmkozh: Are we winning anything by adding `const_` here (as it's already inferred from the type name)?

xdr-generator-rust/generator/src/generator.rs
```diff
@@ -443,21 +578,153 @@ impl RustGenerator {
...
             .collect()
     }
 }
+
+// =============================================================================
+// Borrow analysis
+// =============================================================================
+
+/// Borrow analysis over a whole spec, resolving type references by name.
+struct BorrowAnalysis<'a> {
```
dmkozh, line 798: This seems really clever, though I wouldn't claim I understand all the logic. But I'm not sure if we're winning much by making some types const-compatible, while using separate *Const types for others. Is there some benefit to that?

dmkozh (commented):
On the fuzzing note, I suppose you could create an adapter between const/non-const types via macro or generated code. But that's probably going to be quite complicated.

xdr-generator-rust/generator/src/generator.rs
leighmcculloch: The main benefit is reducing the amount of code that needs compiling that has a direct effect on the developer experience, because doubling the size of this library significantly increases build time, increases memory requirements to a successful build, and because these are used by the soroban-sdk effects every contract developer.

I'll do a test where I remove this and see how bad the compile times are vs keeping it. I'll post back here once I have results.

src/generated/account_entry.rs
leighmcculloch: Ugh that's a bug, will fix. Originally the proposal was a Borrow/Ref type that was independent from the const feature which is why the definition sat separately. But after working through example code and trying to use it a variety of ways I realised that the Ref/Borrow type was almost useless outside of const/static definitions.

leighmcculloch: I researched the best way to name this. Suffixing is common behaviour in the Rust ecosystem which is why I went with that. Originally I used Ref, but that term isn't completely consistent with how other crates use "Ref" because it isn't technically all the borrowed values it is just borrowed values of the heap components, and then I considered "View" which there are some examples of folks using for this type of alternative partial-Ref version of a type, but I want to reserve the term View for when we potentially offer generated sub-definitions of types for partial decoding.

In the end because of https://github.com/stellar/rs-stellar-xdr/pull/580#discussion_r3983763112 I settled on making the types fixed 'static lifetime and so naming them with a Static or Const suffix is more appropriate. I didn't go with Static because it implies a use case that isn't what they're designed for.

Const seems most appropriate. The types, including their functions, are designed for const usage which means their APIs and impl need to take a certain weird shape.

leighmcculloch: We could move the const types into a `const` submodule. It would reduce the noise in the shared namespace.

leighmcculloch: ee1136e

dmkozh: > We could move the const types into a const submodule. It would reduce the noise in the shared namespace.

Yeah, if moving these to a separate module is simple enough, then it's the best option.

src/lib.rs
```diff
@@ -69,6 +69,13 @@
 //! 6. `hex` – Enables support for hex in string representations of some types.
 //! Automatically enabled when serde is enabled.
 //! 7. `schemars` – Enables support for JSON Schema generation. (Experimental)
+//! 8. `const` – Enables the `const` module, containing a form of every type
+//! that can be constructed and encoded to XDR in a const context. Types that
+//! own heap data have a borrowing counterpart in the module that wraps
+//! `&'static` slices instead; the rest are re-exports of the owned types.
+//! Every type in the module has `const_xdr_len` and `const_to_xdr` methods,
+//! so XDR can be encoded at compile time into a fixed-size array. Requires no
```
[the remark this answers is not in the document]
leighmcculloch: Correct, but typedef's of builtin types are already handled by the building writers of the ConstWriter, so this is okay. I don't think this comment should be considered inaccurate because four of the types don't have these functions. Most do, and those four types are essentially renames of primitive types. So I'm not going to change anything here.

xdr-generator-rust/generator/src/generator.rs
leighmcculloch: Adding in all the types makes the crate take 18% longer to compile. So I suggest we keep this.

Makefile
```diff
@@ -27,6 +29,35 @@ readme:
 		| jq -r '"# stellar-xdr\n\n" + .index[.root|tostring].docs' \
 		> README.md
 
+# Run the fuzz corpus.
+# Leak detection is disabled because the fuzzer tests the const impls that
+# contain &'static data and so their Arbitrary impls explicitly leak.
+fuzz:
+	ASAN_OPTIONS=detect_leaks=0 cargo +nightly fuzz run spec-entry -- -rss_limit_mb=0 -max_len=10240 -runs=0
```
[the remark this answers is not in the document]
leighmcculloch: 0 is what its cmmonly used to run once, I'm not sure waht copilit os referring to here

dmkozh (approved):

PR #585, opened 2026-09-16
Fix the publish-dry-run ci

leighmcculloch:
### What

Run the `publish-dry-run` for the x86_64 target as one job per feature combination, instead of one job for all feature combinations, reusing the powerset expansion the prepare job already computes for the `build` and `test` jobs.

### Why

Verifying all 48 feature combinations on a single runner accumulates every combination's build artifacts and disk space is being exhausted on runs on the main branch. For example:
- [run 35022584362](https://github.com/stellar/rs-stellar-xdr/actions/runs/35022584362) failed with `No space left on device` at [combination 45/48](https://github.com/stellar/rs-stellar-xdr/actions/runs/35022584362/job/104561715153) on e0d289a.


.github/workflows/rust.yml
```diff
@@ -156,22 +161,36 @@ jobs:
...
-          cargo-hack-feature-options: --feature-powerset --exclude-features default --group-features base64,serde,serde_json,schemars,arbitrary,hex,rand,type_enum,serde_ignored,test_feature
+        partition: ${{ fromJSON(needs.prepare.outputs.publish-partitions) }}
     uses: stellar/actions/.github/workflows/rust-publish-dry-run.yml@main
     with:
-      runs-on: ${{ matrix.sys.os }}
-      target: ${{ matrix.sys.target }}
-      cargo-hack-feature-options: ${{ matrix.sys.cargo-hack-feature-options }}
+      runs-on: ubuntu-latest
+      target: x86_64-unknown-linux-gnu
+      cargo-hack-feature-options: --feature-powerset --exclude-features default --group-features base64,serde,serde_json,schemars,arbitrary,hex,rand,type_enum,serde_ignored,test_feature --partition ${{ matrix.partition }}
```
leighmcculloch: Re the CodeQL comment on this line ("Workflow does not contain permissions"):

This is pre-existing, not introduced here — `rust.yml` has never had a `permissions` block, including on `main`. It only surfaced now because this PR's diff touches these lines.

`socket-scan.yml`, `regenerate-xdr.yml` and `xdr-generator-rust.yml` do set one, so adding `permissions: contents: read` here would follow the repo's convention. Leaving it out of this PR because it's unrelated to the disk-exhaustion fix and wants checking against the `stellar/actions` reusable workflows this file calls (which inherit the caller's permissions). Happy to do it as a follow-up.

---

.github/workflows/rust.yml
```diff
@@ -156,22 +160,37 @@ jobs:
...
+        feature-args: ${{ fromJSON(needs.prepare.outputs.publish-feature-args) }}
+    uses: stellar/actions/.github/workflows/rust-publish-dry-run-v2.yml@main
     with:
-      runs-on: ${{ matrix.sys.os }}
-      target: ${{ matrix.sys.target }}
-      cargo-hack-feature-options: ${{ matrix.sys.cargo-hack-feature-options }}
+      runs-on: ubuntu-latest
+      target: x86_64-unknown-linux-gnu
+      cargo-hack-feature-options: ''
+      cargo-hack-package-options: ${{ matrix.feature-args }}
```
leighmcculloch, line 196: I needed to use cargo-hack-package-options to place all the options after the 'package' subcommand because the options generated has `--no-default-features` which has to appear after it.

I suspect the fact that cargo-hack-feature-options appears before the 'package' in the command that the workflow runs is a bug. I will consider changing the shared dry run workflow separately.

sisuresh (approved):

